Issue #4, opened 2016-06-27
A user should be able to see pinned Lists on the homepage


briankung:
- [x] The root url displays a page
- [x] Users can sign up, sign in, register
- [x] Users can pin lists
- [x] Users see pinned lists on their homepage

briankung:
Working on this, btw.

Realized recently that I need users in order to have user home pages. I could probably fudge it by storing the homepage data in the user's sessions or whatnot, but that's kind of a shitty solution.

PR #8, opened 2016-07-05
Users

briankung:
Random fixes + homepage + adds users. We need users in order to have users have pinned lists (#4)


briankung:
Steps to test:
1. `docker-compose build`
2. `docker-compose run app rails db:migrate`
3. `docker-compose up`
4. Check that root url returns something - lists
5. Check that you can sign up, sign in, and sign out

PR #9, opened 2016-07-06
Homepage + pinned lists

### LINKED ISSUE #4: A user should be able to see pinned Lists on the homepage

briankung:
Closes #4

This is pretty close - as in, it's functional. That said, it needs some serious refactoring, so I'm submitting a PR so we can leave comments on it and look at the diff more easily.
- [x] Think of a better way to do the homepages lists as per DHH https://github.com/briankung/projectcredo/pull/9#issuecomment-230933651
- [x] Think about how to avoid data inconsistency issues in development with the homepage. Probably shouldn't be an issue... https://github.com/briankung/projectcredo/pull/9#discussion_r69826654
- [x] Think about "Unpin from homepage" wording https://github.com/briankung/projectcredo/pull/9#discussion_r69826684
- [x] See about removing the `current_user` nonsense https://github.com/briankung/projectcredo/pull/9#discussion_r69827121


app/models/user.rb
```diff
@@ -3,4 +3,7 @@ class User < ApplicationRecord
   # :confirmable, :lockable, :timeoutable and :omniauthable
   devise :database_authenticatable, :registerable,
          :recoverable, :rememberable, :trackable, :validatable
+
+  has_one :homepage, dependent: :destroy
+  after_create -> { self.create_homepage }
```
briankung, line 8: This makes me nervous. We're depending on a homepage to exist, but due to testing, it might not. I'm not entirely sure how I want to handle this.

app/views/homepage/show.html.erb
```diff
@@ -19,6 +19,7 @@
         <td><%= link_to 'Show', list %></td>
         <td><%= link_to 'Edit', edit_list_path(list) %></td>
         <td><%= link_to 'Destroy', list, method: :delete, data: { confirm: 'Are you sure?' } %></td>
+        <td><%= link_to 'Unpin from homepage', homepages_list_path(list), method: :delete %></td>
```
briankung: I wonder if there's better phrasing for this

app/views/lists/index.html.erb
```diff
@@ -19,6 +22,13 @@
         <td><%= link_to 'Show', list %></td>
         <td><%= link_to 'Edit', edit_list_path(list) %></td>
         <td><%= link_to 'Destroy', list, method: :delete, data: { confirm: 'Are you sure?' } %></td>
+        <% if current_user%>
+          <% if current_user.homepage.lists.include? list %>
```
briankung: We should probably try to avoid this kind of thing where possible - the `current_user` logic in the views, I mean.

Also pretty sure we can figure out what lists are on the homepage in the controller.

briankung:
Keep this in mind for the HomepageLists, too: http://jeromedalbert.com/how-dhh-organizes-his-rails-controllers/

briankung:
@wsf246 steps to test, should you want to:
1. `docker-compose build`
2. `docker-compose run app rails db:migrate`
3. `docker-compose up`
4. Check that root url redirects to lists when not signed in
5. Check that root url redirects to pinned lists when signed in
6. Check that you can pin and unpin lists from the homepage

app/views/homepage/show.html.erb
briankung: It's temporary, anyway. `¯\_(ツ)_/¯`

wsf246: do my comments appear if I reply all?  Let's see.

wsf246: yasss

app/models/user.rb
briankung, line 8: NAH

app/views/lists/index.html.erb
briankung: Mitigated it.

briankung:
I'm taking a long time with this mostly because I'm realizing that adding routes for the `homepages_lists` is essentially an API that we might want to expose to customers later on.

That being said, it might be prudent to make the change to the route and nothing else.

briankung:
I think that's about it for this revision. I feel like landing on `pins/` as a resource route was really the most major decision. Just like how Twitter has `friendships/` for managing relationships.

The rest is kind of cosmetic.

briankung:
Feel free to test and merge

wsf246:
Looks like there are merge conflicts, probably from me merging the other references branch first.  I'll take a look at the conflicts and see if I cna fix.

briankung:
I'll handle it

wsf246:
youre the best.

PR #10, opened 2016-07-08
Listspapers

wsf246:


config/routes.rb
```diff
@@ -2,6 +2,10 @@
   devise_for :users
   root to: 'homepage#show'
   resources :papers
-  resources :lists
+  resources :lists do
+    member do
+      resources :lists_papers, only: [:create, :destroy]
```
briankung: The same comment I had on my own PR about treating this like an API counts here, too. We will need to clean this API up around the time we switch to a real frontend.

db/schema.rb
```diff
@@ -23,6 +23,19 @@
     t.datetime "updated_at", null: false
   end
 
+  create_table "homepages", id: :uuid, default: -> { "uuid_generate_v4()" }, force: :cascade do |t|
+    t.uuid     "user_id"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+  end
+
+  create_table "homepages_lists", id: false, force: :cascade do |t|
```
briankung: Maybe delete this from the schema for now

app/views/lists/show.html.erb
```diff
@@ -10,5 +10,26 @@
   <%= @list.description %>
 </p>
 
+<ul>
+  <% @list.papers.each do |paper| %>
+    <li>
+      <%=paper.title%>
+    </li>
```
briankung, line 17: Really minor style comment, but this doesn't need to be nested (on different lines)

app/views/lists/show.html.erb
```diff
@@ -10,5 +10,26 @@
...
+<%= form_for(@reference) do |f| %>
+  <div><%= hidden_field_tag :list_id, @list.id %></div>
+
+
+  <div class="field">
+    <%= f.label :paper_id %>
+    <%= f.text_field :paper_id %>
+  </div>
+
+    <%= f.submit "Add Paper", class: "btn btn-primary" %>
```
briankung, line 30: Another style comment, but this should be in line with the `div`. It's a sibling element

briankung:
LGTM, feel free to merge unless you have other issues you want to address

PR #19, opened 2016-07-12
Pubmed json

wsf246:


wsf246:
This creates a module that lets you input search terms and get metadata for 20 papers

wsf246:
there might be an error from nil results, let me check before you merge.

wsf246:
looks like it will just return an empty hash, which is fine.  let me know if you have any questions.

briankung:
Another file naming convention would be to name it `pubmed_search.rb`

wsf246:
done, if it looks good, we can merge

briankung:
Oh yeah, try not to merge your own stuff. I would have had some comments on
that.

wsf246:
sorry, was trying to see if I could get the module to work with a view I'm
making but didn't want to not have all the changes we had just merged into
master.  I'll slow my roll.

PR #27, opened 2016-07-20
Nest author for paper

wsf246:
I'm working on edit/update for this.  There are a few things that we'll need to consider.  The first is adding  how to remove and edit an author.  For example, if I remove an author from a paper, it shouldn't destroy the author.  If I edit, should it create a new author or edit the existing one.

I'm wondering if we have to shift to a has_many through association.  There is some consensus that this is what Rails 4 wants you to do.

### DIFF
--- a/app/controllers/papers_controller.rb
+++ b/app/controllers/papers_controller.rb
@@ -16,6 +16,7 @@ class PapersController < ApplicationController
   # GET /papers/new
   def new
     @paper = Paper.new
+    # SHAME SHAME SHAME DINGDINGDING SHAME SHAME SHAME DINGDINGDING
     5.times { @paper.authors.build }
   end
 
@@ -31,7 +32,6 @@ class PapersController < ApplicationController
     
     respond_to do |format|
       if @paper.save
-
         format.html { redirect_to @paper, notice: 'Paper was successfully created.' }
         format.json { render :show, status: :created, location: @paper }
       else
--- a/app/controllers/papers_controller.rb
+++ b/app/controllers/papers_controller.rb
@@ -16,6 +16,7 @@ class PapersController < ApplicationController
   # GET /papers/new
   def new
     @paper = Paper.new
+    @authors = @paper.authors.build
   end
 
   # GET /papers/1/edit
@@ -70,6 +71,6 @@ class PapersController < ApplicationController
 
     # Never trust parameters from the scary internet, only allow the white list through.
     def paper_params
-      params.require(:paper).permit(:title, :abstract, :link, :doi, :pubmed_id, :published_at, :journal)
+      params.require(:paper).permit(:title, :abstract, :link, :doi, :pubmed_id, :published_at, :journal, authors_attributes: [:name])
     end
 end
--- a/app/controllers/papers_controller.rb
+++ b/app/controllers/papers_controller.rb
@@ -16,7 +16,7 @@ class PapersController < ApplicationController
   # GET /papers/new
   def new
     @paper = Paper.new
-    @authors = @paper.authors.build
+    @authors = 5.times {@paper.authors.build}
   end
 
   # GET /papers/1/edit
@@ -26,10 +26,12 @@ class PapersController < ApplicationController
   # POST /papers
   # POST /papers.json
   def create
-    @paper = Paper.new(paper_params)
 
+    @paper = Paper.new(paper_params)
+    
     respond_to do |format|
       if @paper.save
+
         format.html { redirect_to @paper, notice: 'Paper was successfully created.' }
         format.json { render :show, status: :created, location: @paper }
       else
@@ -71,6 +73,7 @@ class PapersController < ApplicationController
 
     # Never trust parameters from the scary internet, only allow the white list through.
     def paper_params
-      params.require(:paper).permit(:title, :abstract, :link, :doi, :pubmed_id, :published_at, :journal, authors_attributes: [:name])
+      params.require(:paper).permit(:title, :abstract, :link, :doi, :pubmed_id, :published_at, :journal, 
+        authors_attributes: [:name])
     end
 end
--- a/app/controllers/papers_controller.rb
+++ b/app/controllers/papers_controller.rb
@@ -16,7 +16,7 @@ class PapersController < ApplicationController
   # GET /papers/new
   def new
     @paper = Paper.new
-    @authors = 5.times {@paper.authors.build}
+    5.times { @paper.authors.build }
   end
 
   # GET /papers/1/edit
@@ -63,7 +63,7 @@ class PapersController < ApplicationController
       format.html { redirect_to papers_url, notice: 'Paper was successfully destroyed.' }
       format.json { head :no_content }
     end
-  end
+  end 
 
   private
     # Use callbacks to share common setup or constraints between actions.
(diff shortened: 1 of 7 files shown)


wsf246:
Write now I can add authors to papers in create that will be associated to the paper.  I still need to find a way to add an existing author to a paper.  I think it should work by ignoring an author that has already been created and just add the author to the paper on the paper's creation.

briankung:
I'm a bit ambivalent about `has_many _, through: :`. `has_and_belongs_to_many` is really simple, which I like. On the other hand, `has_many through:` does seem to be the convention.

briankung:
I want to stick with the simplest option until it no longer works - basically when we want to store information about the relationship on the relationship, itself.

However, if you can convert it with no problems, I wouldn't be against switching. I just don't know that there's a justification for it - or whether the problem(s) you're encountering are related to the type of relationship we're using.

app/controllers/papers_controller.rb
```diff
@@ -16,6 +16,7 @@ def show
   # GET /papers/new
   def new
     @paper = Paper.new
+    @authors = @paper.authors.build
```
briankung: This results in one author, btw

app/models/author.rb
```diff
@@ -1,3 +1,5 @@
 class Author < ApplicationRecord
   has_and_belongs_to_many :papers
+  validates :name, uniqueness: { case_sensitive: false }, presence: true
+
```
briankung: Let's not add whitespace unnecessarily

app/models/paper.rb
```diff
@@ -2,4 +2,7 @@ class Paper < ApplicationRecord
   has_and_belongs_to_many :authors
   has_many :lists, through: :references
   has_many :references
+
+  accepts_nested_attributes_for :authors
+  validates_associated :authors
```
briankung, line 7: Neat, so that's how it works. I didn't remember.

app/controllers/papers_controller.rb
```diff
@@ -16,6 +16,7 @@ def show
   # GET /papers/new
   def new
     @paper = Paper.new
+    @authors = 5.times {@paper.authors.build}
```
briankung: I'm pretty sure this results in `@authors = 5`, though it may or may not associate 5 empty authors to the paper. What's your intention with this? To have 5 form fields show up?

app/models/paper.rb
```diff
@@ -2,4 +2,7 @@ class Paper < ApplicationRecord
   has_and_belongs_to_many :authors
   has_many :lists, through: :references
   has_many :references
+
+  accepts_nested_attributes_for :authors, reject_if: proc { |attributes| attributes['name'].blank? }
```
briankung, line 6: Shouldn't this be handled by the authors validations?

briankung:
> I think it should work by ignoring an author that has already been created and just add the author to the paper on the paper's creation.

I agree. I'm not seeing where you're adding an author to a paper, though. It seems to be handled by Rails magic at the moment?

briankung:
Seems so: https://stackoverflow.com/questions/3579924/accepts-nested-attributes-for-with-find-or-create

Looks like it's handled by `autosave_associated_records_for_author`. You could override this in combination with `find_or_create_by` http://apidock.com/rails/v4.2.1/ActiveRecord/Relation/find_or_create_by

app/models/author.rb
wsf246: sorry, I have an undisciplined way of writing code.  I'll add and get rid of things and forget I had white space.

app/controllers/papers_controller.rb
wsf246: yea, this will add 5 instances of the authors text field.  I'm also wondering how we want to handle title.  For now I'm ignoring it because I feel like its not essential.  It might even be easier later on if we delimit titles from the names later on from name.

app/models/paper.rb
wsf246, line 6: I believe when I read the tutorial, this validation was needed here for nested attributes.  I'll double cheeck, but its meant to not create blank authors.  Ie, I have 5 author fields and if I enter 4 names, only 4 will be created.

wsf246:
yup, rails magic.  I actually had it working yesterday but I had <% author %> in my show yesterday instead of <%= author %> so I thought it wasn't associating.  I was wrong. :D

app/controllers/papers_controller.rb
briankung: OK. Why don't we do this:

``` ruby
5.times { @paper.authors.build } # Note spaces inside brackets XD
@authors = @paper.authors # Though you should be able to reach all of them through the @paper anyway
```

app/models/paper.rb
briankung, line 6: I see. It might silently fail if this isn't here? Like the authors creation will fail, but the paper will save or something.

app/controllers/papers_controller.rb
wsf246: Is there a reason to split it into 2 lines?  I just want to make sure I understand the logic for future reference.  Easier to read?

briankung: Well...you don't even need the 2nd line to start with, so, no. The only line that matters is the `5.times { @paper.authors.build }`

The second line is just to capture the `@authors` variable in case you're using it in a view. But really, you shouldn't do that, because you can always get the authors from `@paper.authors`

briankung: Also, like I mentioned earlier, `5.times { @paper.authors.build }` actually evaluates to `5`.

```
irb(main):002:0> 1 + 5.times {print 'hey'}
heyheyheyheyhey=> 6
```

briankung: So `@authors = 5.times { @paper.authors.build }` stores `5` into the `@authors` instance variable

wsf246: gotcha.  I'll just put 5 where the variable is sitting in the form then.
I'll see if I run into any problems.

briankung:
Is this still WIP?

wsf246:
Nope.  If it LGTM, we should do it

app/controllers/papers_controller.rb
```diff
@@ -25,10 +26,12 @@ def edit
   # POST /papers
   # POST /papers.json
   def create
-    @paper = Paper.new(paper_params)
 
+    @paper = Paper.new(paper_params)
+    
     respond_to do |format|
       if @paper.save
+
```
briankung: Whitespace

app/controllers/papers_controller.rb
```diff
@@ -16,6 +16,7 @@ def show
   # GET /papers/new
   def new
     @paper = Paper.new
+    5.times { @paper.authors.build }
```
briankung, line 20: This is a hack that we should probably mark with some sort of comment like TODO or SHAME. We'll be able to remove this when we have a real front-end.

app/models/paper.rb
briankung, line 6: Bump. Will this silently fail?

briankung:
Once we resolve the minor things here, you can merge, methinks

briankung:
If this closes an issue, please make sure to close that issue as well. And delete the branch plz. Thanks!

PR #28, opened 2016-07-23
Pubmed support

### LINKED ISSUE #6: When adding a paper by DOI, additional information should be pulled from a third party source
Probably Pubmed to grab title, authors, etc.

briankung:
Closes #6

### DIFF
--- a/app/controllers/references_controller.rb
+++ b/app/controllers/references_controller.rb
@@ -12,11 +12,17 @@ class ReferencesController < ApplicationController
       pubmed = Pubmed.new
       results = pubmed.search paper_id
       data = results['result'][paper_id]
-      paper = Paper.create(
-        pubmed_id: data['uid'],
-        title: data['title'],
-        published_at: data['pubdate'],
-        doi: data['elocationid'].sub(/^doi: /, "")
+      paper = (
+        Paper.find_by(pubmed_id: paper_id) ||
+        Paper.find_by(doi: paper_id) ||
+        # I really hate this due to long experience with it,
+        # but we should probably some sort of mapping
+        Paper.create(
+          pubmed_id: data['uid'],
+          title: data['title'],
+          published_at: data['pubdate'],
+          doi: data['elocationid'].sub(/^doi: /, "")
+        )
       )
     end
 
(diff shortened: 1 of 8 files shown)


briankung:
Don't merge until this method gets looked at: https://github.com/briankung/projectcredo/pull/28/files#diff-6feb4b64b85f4f775a2b7af90d544a3aR4

app/controllers/references_controller.rb
```diff
@@ -1,10 +1,37 @@
...
+      paper = (
+        Paper.find_by(pubmed_id: paper_id) ||
+        Paper.find_by(doi: paper_id) ||
+        # I really hate this due to long experience with it,
+        # but we should probably some sort of mapping
+        Paper.create(
+          pubmed_id: data['uid'],
+          title: data['title'],
+          published_at: data['pubdate'],
+          doi: data['elocationid'].sub(/^doi: /, "")
```
wsf246, line 25: The other piece we'll need to add is authors.  i'm not sure if there's rails magic that handles this, but if there's an author name that is already in the db, it should map to the existing uuid.

wsf246, line 25: I can take a stab at that?

briankung, line 25: Sure. I would, but I'm stuck on the highway.

Hurry. The zombies are coming.

wsf246, line 25: I'm playing around with it and it's super cool.  I think we should add some
basic clean up of entry like removing blank spaces.  I'm also getting an
error for trying to add a doi.  So i think we should add a "No paper found"
error message.  I'll make these updates along with adding the author.

briankung, line 25: Error handling can be pretty nasty. Also please don't do anything clever
until we refactor that method.

briankung, line 25: Basically don't add unnecessary things. Dealing with the complexity
afterwards is much worse than having a slightly broken feature.

wsf246, line 25: I'll take it as a compliment that you think i can do anything clever.

wsf246:
So I'm thinking I merge this so I can work off of the author stuff that we already merged as well.  Thoughts?

wsf246:
We can leave the issue open as I work through it.

briankung:
Sure, can you write an additional GitHub issue to refactor
ReferencesController#create ?

PR #46, opened 2016-08-04
Papers partial

briankung:
Reuses the papers index on the list show page


briankung:
screen shot 2016-08-04 at 6 36 04 pm

PR #48, opened 2016-08-08
Voting (closes #47)

### LINKED ISSUE #47: A user should be able to upvote lists and papers within a list

briankung:
Lists can be upvoted, but not references, yet. It's a bit tricky. Also, the CSS / HTML is clearly lacking.


briankung:
Still kinda WIP due to CSS and what not, but voting is done for the most part

PR #49, opened 2016-08-08
Create bias tags

wsf246:


app/views/papers/_form.html.erb
```diff
@@ -53,6 +53,11 @@
     <%= f.text_field :tag_list, value: paper.tag_list.join(",") %>
   </div>
 
+  <div class="field">
+    <%= f.label :bias_list, 'Biases (the following biases are accepted, please seperate with a comma: selection, performance, detection, attrition, reporting, small sample size)' %>
```
briankung: Should be "separate"

app/models/paper.rb
```diff
@@ -1,10 +1,19 @@
...
 
   has_and_belongs_to_many :authors
   has_many :lists, through: :references
   has_many :references, dependent: :destroy
 
   accepts_nested_attributes_for :authors, reject_if: proc { |attributes| attributes['name'].blank? }
   validates_associated :authors
+  validate :bias_list_inclusion
+
+  def bias_list_inclusion
```
briankung: Let's rename this something more English-like, like `validate :allowed_biases`

app/models/paper.rb
```diff
@@ -1,10 +1,19 @@
...
+    bias_list.each do |bias|
+      errors.add(bias,"is not valid") unless %w(selection performance detection attrition reporting small\ sample\ size).include?(bias)
```
briankung: Let's break `%w(selection performance detection attrition reporting small\ sample\ size)` into its own private method, perhaps called `valid_biases`. Then you could do:

``` ruby
def allowed_biases
  invalid_biases = bias_list - valid_biases
  invalid_biases.each {|b| errors.add(b, 'is not valid') }
end
```

config/application.rb
```diff
@@ -14,5 +14,7 @@ class Application < Rails::Application
 
     # For adding subdirectories
     config.autoload_paths += Dir[Rails.root.join('app', 'models', '{**}')]
+  	# Force lowercase for Acts on Taggable
+  	ActsAsTaggableOn.force_lowercase = true
```
briankung: Let's put all of these in its own initializer file called `acts_as_taggable_on.rb`

app/models/paper.rb
```diff
@@ -1,10 +1,31 @@
 class Paper < ApplicationRecord
   acts_as_taggable
+  acts_as_taggable_on :biases, :methodologies
+  ActsAsTaggableOn.strict_case_match = true
```
briankung: Ditto here wrt the initializer file

wsf246:
closes #50 and #51

app/models/paper.rb
```diff
@@ -1,10 +1,31 @@
...
   validates_associated :authors
+  validate :allowed_biases, :allowed_methodologies
+
+  def allowed_biases
+	  invalid_biases = bias_list - valid_biases
+	  invalid_biases.each {|b| errors.add(b, 'is not an accepted bias') }
+	end
+
+  def valid_biases
+    %w(selection performance detection attrition reporting small\ sample\ size)
```
briankung: To make it easier to edit these in the future, it might be simpler to break these into separate lines and just use the normal array syntax:

``` ruby
[
  'selection',
  'performance',
  'detection',
  'attrition',
  'reporting',
  'small sample size'
]
```

briankung: Same with the valid methodologies. Also, putting them near each other would help with readability.

app/views/papers/_form.html.erb
```diff
@@ -53,6 +53,16 @@
     <%= f.text_field :tag_list, value: paper.tag_list.join(",") %>
   </div>
 
+  <div class="field">
+    <%= f.label :methodology_list, 'Methodologies (the following methodologies are accepted, please separate with a comma: meta-analysis, systematic review, randomized control trial, quasi-experiment, cohort, case-control, cross-sectional survey, case report)' %>
```
briankung: Later when we force auto-complete on the frontend, we can ensure they can't input anything else. Right now this is fine.

briankung: Wait...why not make these multi-select check boxes?

briankung: screen shot 2016-08-09 at 7 16 46 pm

wsf246: multi select checkboxes is prob the best interim solution.  I'm hoping in
the long run, we can use tag boxes that you can add and remove on the show
page.  Like stackoverflow:

[image: Inline image 1]

app/models/paper.rb
wsf246: when you say near each other?

briankung: I mean, put them next to each other.

``` ruby
def valid_biases
  # ...
end

def valid_methodologies
  # ...
end
```

wsf246: ah ok

PR #58, opened 2016-08-16
Adds RailsAdmin

briankung:


briankung:
I'm going to close this because I don't thin it's useful right now, but I'll leave the branch up in case we want to include it later.

PR #59, opened 2016-08-17
Paper index css

wsf246:
Pretties up the Paper Index

### DIFF
--- a/app/views/papers/index.html.erb
+++ b/app/views/papers/index.html.erb
@@ -2,7 +2,7 @@
 
 <h1>Papers</h1>
 
-<%= link_to 'New Paper', new_paper_path %>
+<%= link_to 'New Paper', new_paper_path %> </br>
 
 <%= render partial: 'papers/index', locals: { papers: @papers } %>
 
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -39,4 +39,5 @@ body {
 
 .paper-title {
     margin: 0;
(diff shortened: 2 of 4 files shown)


app/views/papers/index.html.erb
```diff
@@ -2,7 +2,7 @@
 
 <h1>Papers</h1>
 
-<%= link_to 'New Paper', new_paper_path %>
+<%= link_to 'New Paper', new_paper_path %> </br>
```
briankung, line 5: This irritates me a little bit, but I did something equally annoying with the inline style elsewhere.

briankung:
Is it supposed to look like this?

screen shot 2016-08-17 at 12 07 45 pm

https://afternoon-meadow-22528-pr-59.herokuapp.com/lists/98248c4e-b5b0-4e29-94bf-c68f06f9e3c0

briankung:
I thought the Abstract, Details, etc. would have a darker background.

briankung:
Otherwise, LGTM.

wsf246:
Yea I haven't been able to get the bootstrap to do exactly what I want.
Basically there's a JavaScript tab function but you can't close what's
shown after you open it which is really annoying.  I'm gonna play around w
it more today to see if I can find a better way, but yes it should be
darker and it should only show each item at once.

wsf246:
Also add some mythologies bias and tags to see if you're OK w the way it
looks. We can change the colors up but I didn't want it to be overly
distracting

PR #76, opened 2016-09-13
Add usernames

### LINKED ISSUE #71: Users should have usernames
This will be particularly important for comments

wsf246:
I'm going to do some quick formatting w/ bootstrap form stuff for the devise views, but other than that it seems to work.  I'm like 85% lucid of what this code is doing and i used this tutorial:
https://github.com/plataformatec/devise/wiki/How-To:-Allow-users-to-sign-in-using-their-username-or-email-address

Hopefully I didn't break anything.

Closes #71


app/models/user.rb
```diff
@@ -4,10 +4,40 @@ class User < ApplicationRecord
...
+  # Virtual attribute for authenticating by either username or email
+  # This is in addition to a real persisted field like 'username'
+  attr_accessor :login
+
+  validates_format_of :username, with: /^[a-zA-Z0-9_\.]*$/, :multiline => true
+  validates :username,
+  :presence => true,
+  :uniqueness => {
+    :case_sensitive => false
+  }
```
briankung: Let's use Ruby 1.9+ style hashes where possible

briankung: So `presence: true` and the likes

wsf246:
done!

app/views/users/mailer/confirmation_instructions.html.erb
```diff
@@ -0,0 +1,5 @@
+<p>Welcome <%= @email %>!</p>
```
briankung: This should probably be the username now

app/views/users/mailer/password_change.html.erb
```diff
@@ -0,0 +1,3 @@
+<p>Hello <%= @resource.email %>!</p>
```
briankung: This should also be the username. Or we should remove it. I kind of like the idea of removing it because the less information we give a potential attacker, the better.

briankung: But I could go either way.

app/views/users/mailer/confirmation_instructions.html.erb
briankung: I may have changed my mind. If we give them the username, they now have the email address as well as the username. Let's either keep it as email or remove it entirely with a message like "Welcome! Someone signed up for ... with this email address. If it wasn't you, just ignore it."

app/models/user.rb
```diff
@@ -4,10 +4,40 @@ class User < ApplicationRecord
...
   acts_as_voter
+
+  def self.find_for_database_authentication(warden_conditions)
+    conditions = warden_conditions.dup
+    if login = conditions.delete(:login)
+      where(conditions.to_h).where(["lower(username) = :value OR lower(email) = :value", { :value => login.downcase }]).first
+    elsif conditions.has_key?(:username) || conditions.has_key?(:email)
+      conditions[:email].downcase! if conditions[:email]
+      where(conditions.to_h).first
+    end
```
briankung, line 28: I'm not sure what this `elsif` block handles. It looks like it checks for a `username` or `email` if `login` isn't provided, but the only authentication key that you can use to log in is `login`.

briankung, line 28: Oh, this is at a model level, not for the forms. So `login` remains the only parameter you can authenticate by.

But it does look like it downcases the email and not the username, if one is provided. I would want to downcase the username as well, but I'm not sure when and where `find_for_database_authentication` is used.

app/models/user.rb
```diff
@@ -4,10 +4,40 @@ class User < ApplicationRecord
...
+  end
+
+  before_save do
+    self.email.downcase! if self.email
+  end
+
+  def self.find_for_authentication(conditions)
+    conditions[:email].downcase!
```
briankung: I just noticed that this also doesn't handle usernames. Can you try to figure out when, where, and why `find_for_database_authentication` and `find_for_authentication` are used? There's a lot of metaprogramming in the Devise library (to support arbitrary custom models like User, Admin, Customer, whatever you want to name your model), so it's going to be kind of difficult. But if you can figure out the reasoning for having the two finders, it would help a lot.

app/views/users/mailer/reset_password_instructions.html.erb
```diff
@@ -0,0 +1,8 @@
+<p>Hello <%= @resource.email %>!</p>
```
briankung: You missed a few of the mailer email mentions

app/models/user.rb
wsf246: This was what the tutorial said:

Because we want to change the behavior of the login action, we have to overwrite the find_for_database_authentication method. The method's stack works like this: find_for_database_authentication calls find_for_authentication which calls find_first_by_auth_conditions. Overriding the find_for_database_authentication method allows you to edit database authentication; overriding find_for_authentication allows you to redefine authentication at a specific point (such as token, LDAP or database). Finally, if you override the find_first_by_auth_conditions method, you can customize finder methods (such as authentication, account unlocking or password recovery).

I was pretty confused by this code as well.  I would love to do some pair coding if you can walk me through this logic.

app/models/user.rb
briankung: Also, if you _just_ have a `:username` in the conditions hash, then this method will break upon trying to downcase the email

wsf246:
initializers?

briankung:
Ugh, I hate commit comments

config/initializers/devise.rb
```diff
@@ -34,7 +34,7 @@
   # session. If you need permissions, you should implement that in a before filter.
   # You can also supply a hash where the value is a boolean determining whether
   # or not authentication should be aborted when the value is not present.
-  # config.authentication_keys = [:email]
+  config.authentication_keys = [:logins]
```
briankung: This is plural now

briankung: Logins

wsf246: I think that's a typo.  At Costco right now, I will fix it when I come
back.  I was wondering why it broke.

wsf246:
This is the best explanation I saw for the find_for_database_authentication and find_for_authentication.  It looks like the first is just a wrapped version of the second and the second is a wrapped version of find_first_by_auth_conditions.  I'm not sure why they nest these methods within each other, I suppose for different tasks, but it looks like these are just used solely during authentication.

https://www.codetriage.com/doc_methods/1097

app/models/user.rb
```diff
@@ -4,10 +4,41 @@ class User < ApplicationRecord
...
+      conditions[:email].downcase! if conditions[:email]
+      conditions[:username].downcase! if conditions[:username]
+      where(conditions.to_h).first
+    end
+  end
+
+  def self.find_for_authentication(conditions)
+    conditions[:email].downcase! if conditions[:email]
+    super(conditions)
+  end
```
briankung, line 30: Can you try removing this to see if everything works without it? I feel like this is not really doing anything for us, especially since you are downcasing in `find_for_database_authentication`.

app/models/user.rb
```diff
@@ -4,10 +4,36 @@ class User < ApplicationRecord
...
+
   acts_as_voter
+
+  def self.find_for_database_authentication(warden_conditions)
+    conditions = warden_conditions.dup
+    if login = conditions.delete(:login)
+      where(conditions.to_h).where(["lower(username) = :value OR lower(email) = :value", { :value => login.downcase }]).first
+    elsif conditions.has_key?(:username) || conditions.has_key?(:email)
+      conditions[:email].downcase! if conditions[:email]
+      conditions[:username].downcase! if conditions[:username]
```
briankung: This introduces the problem that I can save a `User` with a case sensitive username and then have it not return correctly if I give `#find_for_database_authentication` the proper username.

```
irb(main):003:0> u
=> #<User ... username: "USeR">
irb(main):004:0> User.find_for_database_authentication username: 'USeR'
  User Load (0.8ms)  SELECT  "users".* FROM "users" WHERE "users"."username" = $1 ORDER BY "users"."id" ASC LIMIT $2  [["username", "user"], ["LIMIT", 1]]
=> nil
```

briankung: So we should probably downcase the username before save, too, if we're gonna do this.

briankung (changes requested):
Let's downcase the usernames in a `before_save` hook, too.

app/models/user.rb
wsf246: still works

app/models/user.rb
```diff
@@ -24,16 +24,9 @@ class User < ApplicationRecord
...
     conditions = warden_conditions.dup
     if login = conditions.delete(:login)
-      where(conditions.to_h).where(["lower(username) = :value OR lower(email) = :value", { :value => login.downcase }]).first
-    elsif conditions.has_key?(:username) || conditions.has_key?(:email)
-      conditions[:email].downcase! if conditions[:email]
-      conditions[:username].downcase! if conditions[:username]
-      where(conditions.to_h).first
+      return where(conditions.to_h).where(["username = :value OR lower(email) = :value", { :value => login.downcase }]).first
     end
+    super(warden_conditions)
```
briankung: So, I tested this (by inserting a raise and debugging from there) and it's safer to use `conditions` than it is to use `warden_conditions` because there is no `login` column on the users table.

So I was able to confirm that there was a parent method, and that it required conditions that correlate to the user table.

briankung (approved):

app/models/user.rb
briankung: But it should also basically never get there.

PR #86, opened 2016-09-18
Revert UUIDs

briankung:
This reinstates integer IDs across all relationships and primary keys. In order to use this schema, you would have to use: `docker-compose run app rails db:drop db:create db:schema:load`

Reading:
- https://rclayton.silvrback.com/do-you-really-need-a-uuid-guid

We need to figure out a scheme for easily shareable lists, but I don't think UUIDs are it.

We should also look into friendly ids. Now that users have usernames, we can probably do what github does and namespace a friendly list name under the username. I'm wondering how to handle urls for public lists, as well.


wsf246:
I agree with this strategy.  In my last iteration i used a gem to create
name friendly ids based on the title etc, but I'm ok w/ whichever approach
we end up going with.

https://github.com/norman/friendly_id

briankung:
Also, it turns out that we don't need anything like https://github.com/jalkoby/squasher because we are supposed to use `rake db:schema:load`, anyway

briankung:
Agreed. I was looking at using that gem, as well.

PR #87, opened 2016-09-21
WIP - Votes for comments no uui ds

wsf246:
I'm still getting the same issue after merging:
image

I'll keep looking into it, but it seems like the move away from uuids is causing an issue with the comments and that helper as well.


briankung:
Looks like something in closure tree is screwing up. Maybe some interaction
with acts as votable is triggering an infinite update loop.

I'm almost home, so I'll check it out for a bit.

wsf246:
This branch is being overwritten by votesForComments

PR #91, opened 2016-09-26
Votes for comments

### LINKED ISSUE #83: Reference comments should be ordered by vote count.

wsf246:
**Problem:** Comments have no votes and are sorted purely on time of creation

**Solution:**  Adds votes for comments and sorts comments by votes.  Also includes some styling 

**Complications:**  Does not fully support comments on papers index

closes #83


wsf246:
closes #83

wsf246:
Hey if this LGTM, I'll work on resolving the conflicts.

app/controllers/comments_controller.rb
```diff
@@ -31,7 +31,7 @@ def update
         format.html { redirect_to :back, notice: 'Comment was successfully updated.' }
         format.json { render :show, status: :ok, location: @comment }
       else
-        format.html { render :edit }
+        format.html { render :back }
```
briankung: What does `render :back` do?

app/models/comment.rb
```diff
@@ -2,5 +2,8 @@ class Comment < ApplicationRecord
   belongs_to :user
   belongs_to :commentable, polymorphic: true
 
-  has_closure_tree
+  has_closure_tree order: 'sort_order'
+  acts_as_votable
+
+  validates :content, presence: true
```
briankung, line 8: Nice - should definitely have this

briankung:
Definitely work on resolving the conflicts. I'm still reviewing.

app/helpers/comments_helper.rb
```diff
@@ -1,6 +1,6 @@
 module CommentsHelper
   def comments_tree_for(comments)
-    comments.map do |comment, nested_comments|
+    comments.sort_by{ |k,v| v[:cached_vote_up] }.map do |comment, nested_comments|
```
briankung: Just making sure you're aware that this will only sort the top level comments.

app/views/papers/_index.html.erb
```diff
@@ -52,8 +52,13 @@
         <% end %>
       </small>
       <ul class="nav nav-pills" role="tablist">
+        <% comments_count = paper_reference ? paper_reference.comments.map {|c| c.self_and_descendants.length}.inject(0, :+).to_s : paper.references.map { |r| r.comments.map {|c| c.self_and_descendants.length}.inject(0, :+)}.inject(0, :+).to_s %>
```
briankung: This...is way too long

briankung: RE: Switching based on `paper_reference`, but then providing an alternate: The only reason I switch based on paper reference is to change the display depending on whether it's a on a list or not. It's only a positive - that is, if `paper_reference` exists, I display things. If you display things anyway, you might as well just remove the check for `paper_reference`. `paper.references.map...` will work just as well for both cases.

app/views/comments/_comment.html.erb
```diff
@@ -4,11 +4,15 @@
...
-    <%= render 'comments/form', comment: comment.children.build, display_comment_field: false %>
-  <% end %>
+  <ul>
+    <li><%= render 'votes/toggle_like', votable: comment %></li>
+    <% if current_user %>
+      <li><%= link_to 'edit', edit_comment_path(comment), class: "btn btn-link btn-xs" %></li>
+      <li><%= link_to 'reply', "#", class: 'toggle-reply btn btn-link btn-xs', 'data-toggle-hide-id' => comment.id %></li>
+      <li><%= link_to 'delete', comment_path(comment), class: "btn btn-link btn-xs", method: :delete %><li>
+    <% end %>
+  </ul>
```
briankung, line 14: As much as I dislike adding HTML, I have to admit this looks and works fine.

app/views/papers/_index.html.erb
briankung: ^ We do have to fix that, though - we should not be checking for paper_references in the first place.

briankung:
``` ruby
paper.references.map { |r| r.comments.map {|c| c.self_and_descendants.length}.inject(0, :+)}.inject(0, :+).to_s
```

Why are you doubling up on the injects?

briankung: This is a really db intensive operation. Is there any other way we can do this?

app/views/papers/_index.html.erb
```diff
@@ -52,8 +52,13 @@
...
         <% %i{abstract details comments}.each do |tab| %>
-          <li><%= link_to tab.capitalize, '', "data-detail-id" => "#{tab}-#{index}", class: "small paper-detail-tab" %></li>
+          <% if tab == :comments %>
+            <li><%= link_to tab.capitalize.to_s + "(" + comments_count + ")", '', "data-detail-id" => "#{tab}-#{index}", class: "small paper-detail-tab" %></li>
```
briankung: In the future, please try to avoid string concatenation with `:+`. Mostly because I think it's ugly.

app/views/papers/_index.html.erb
```diff
@@ -52,8 +52,13 @@
...
+          <% else %>
+            <li><%= link_to tab.capitalize, '', "data-detail-id" => "#{tab}-#{index}", class: "small paper-detail-tab" %></li>
+          <% end %>
```
briankung: Avoid writing additional code where possible. It's also kind of redundant in the most complicated part of the line. Instead of:

``` erb
<% if tab == :comments %>
  <li><%= link_to tab.capitalize.to_s + "(" + comments_count + ")", '', "data-detail-id" => "#{tab}-#{index}", class: "small paper-detail-tab" %></li>
<% else %>
  <li><%= link_to tab.capitalize, '', "data-detail-id" => "#{tab}-#{index}", class: "small paper-detail-tab" %></li>
<% end %>
```

You could do something like:

``` erb
<% 
  tab_link_text = tab.capitalize
  tab_link_text += " (#{comment_count})" if tab == :comment
%>
<li><%= link_to tab_link_text, '', "data-detail-id" => "#{tab}-#{index}", class: "small paper-detail-tab" %></li>
```

app/views/votes/_toggle_like.html.erb
```diff
@@ -3,8 +3,15 @@
   liked = user_signed_in? && current_user.voted_for?(votable)
 %>
 
-<%= form_tag vote_path, method: (liked ? :delete : :post) do %>
+<%= form_tag vote_path, method: (liked ? :delete : :post), class: "like-form" do %>
```
briankung, line 6: I was thinking about doing the opposite of adding HTML classes like these, actually. I don't think we should be naming them after their backend models, but their frontend functionality. That way we could reuse them across more views.

That said, I don't care much about this at the moment to force a change.

app/views/votes/_toggle_like.html.erb
```diff
@@ -3,8 +3,15 @@
...
   <%= text_field_tag 'type', votable.class.to_s.downcase, hidden: true %>
-  <%= submit_tag '👍 ' + votable.get_likes.size.to_s, class: "btn btn-default btn-sm pull-left #{"active" if liked}" %>
+  <% if votable.class.to_s == "Comment" %>
+    <%= button_tag(type: 'submit', class: "btn btn-link btn-xs #{"inactive" unless liked}") do %>
+      <%= votable.get_likes.size.to_s %>
+      <span class="glyphicon glyphicon-triangle-top"></span>
+    <% end %>
+  <% else %>
+    <%= submit_tag '👍 ' + votable.get_likes.size.to_s, class: "btn pull-left btn-default btn-sm #{"active" if liked}" %>
+  <% end %>
```
briankung: We should not be switching this based on the votable class. That is an antipattern: https://sourcemaking.com/refactoring/replace-conditional-with-polymorphism

Since this seems to be purely a style difference, please see if you can find a way to display a different glyph using CSS.

app/controllers/comments/votes_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  def destroy
+    current_user.unlike @comment
+    if @comment.parent_id
+      closest_comment_by_votes = @comment.siblings.min_by {|c| (c.cached_votes_up - @comment.cached_votes_up).abs }
+      if @comment.cached_votes_up >= closest_comment_by_votes.cached_votes_up
+        closest_comment_by_votes.prepend_sibling(@comment)
+      else
+        closest_comment_by_votes.append_sibling(@comment)
+      end
```
briankung: Aren't these two blocks the same? You could capture this in a method so you have one definitive source of knowledge for prepending/appending comments based on votes.

app/controllers/comments/votes_controller.rb
```diff
@@ -0,0 +1,41 @@
+class Comments::VotesController < ApplicationController
+  before_action :ensure_current_user
+  before_action :set_comment
+
+  def create
+    current_user.likes @comment
+    if @comment.parent_id
+      closest_comment_by_votes = @comment.siblings.min_by {|c| (c.cached_votes_up - @comment.cached_votes_up).abs }
+      if @comment.cached_votes_up >= closest_comment_by_votes.cached_votes_up
```
briankung: Won't this make it so that it will be ordered by vote, and then newest first, since you're prepending the new comment to comments with similar vote scores?

briankung (changes requested):
Try to write less code, more simply and readably. Especially check yourself when you're writing HTML to change the visual style of an element, which you seem to do quite often. Make sure there isn't a CSS solution for it first.

Also, as uncomfortable as I am with the comment ordering, I do think it might be the best solution we have unless `closure_tree` adds more functionality.

app/views/papers/_index.html.erb
briankung: I misinterpreted this, btw, the `paper_reference` and `paper.references` do different things. My bad

briankung: The cleanest I was able to get this was:

`paper.references.inject(0) { |total, r| total += r.comments.inject(0) {|memo,c| memo += c.self_and_descendants.length}}`

briankung: We could also count the `.comment` divs with JS. Or somehow get the rendered partial to count it. I'm looking into the latter now.

app/controllers/comments_controller.rb
wsf246: I think it was supposed to be redirect_to :back

app/helpers/comments_helper.rb
wsf246: I don't think i need this so I'm removing it.

wsf246: good catch

app/assets/javascripts/comments.js
```diff
@@ -3,4 +3,12 @@ $(document).ready(function() {
...
     $(e.target).parents("ul").next('.comment-reply').toggleClass('hidden');
   });
-});
\ No newline at end of file
+});
+
+
+$(document).ready(function() {
+    $( '[data-detail-id^=comments-]' ).each(function(index) {
+      var comments = $( '#comments-'+index ).find(".comment").length;
```
briankung: This variable should probably communicate the idea of "number of comments" instead of just "comments"

app/views/votes/_toggle_like.html.erb
wsf246: I agree with this, I think we'll have better visibility on this when we make changes based on react

app/views/votes/_toggle_like.html.erb
wsf246: not sure I know the best way to solve this, let's talk about it

briankung: Neither do I, but it's probably something like this https://stackoverflow.com/questions/22261048/using-glyphicon-as-background-image-in-css-rails-4-bootstrap-3

briankung: The answer is kind of garbage, though

briankung: This is kind of better, and along the same lines: https://stackoverflow.com/questions/14238905/using-twitter-bootstrap-glyphicons-in-css

briankung: Before and after psuedo selectors: 

https://developer.mozilla.org/en-US/docs/Web/CSS/::before

https://css-tricks.com/almanac/selectors/a/after-and-before/

briankung:
That glyphicon CSS is so cool

app/controllers/comments/votes_controller.rb
wsf246: I think this is how it works for all the comments under the top comment class.  It will essentially keep sort by votes and then last updated which includes voting activity.  I'm not sure we can do to incorporate more in the sort order than that unless we get more obtrusive with like a "sort order score".  Let me know if you can think of different way though.

briankung: Just have it append if it's equal

wsf246:
The prepend is for the case for when it latches onto something that is
smaller for an upvote.  So 2 comments with a vote each, when you upvote
one, it will need to prepend the other in the sort order to go to the top.

briankung:
I understand that. If they're the same vote level and you APpend it,
though, then the newer comments with the same vote score will be on the
bottom.

Right?

wsf246:
It sorta does that.  If there's multiple comments with the same votes, it'll append it to the one at the top, which is a little weird.  I don't really care whichever way so long as the primary vote order is correct.

app/controllers/comments/votes_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    def insert_comment comment
+      if comment.parent_id
+        closest_comment_by_votes = comment.siblings.min_by {|c| (c.cached_votes_up - comment.cached_votes_up).abs }
+        if comment.cached_votes_up > closest_comment_by_votes.cached_votes_up
+          closest_comment_by_votes.prepend_sibling(comment)
+        elsif comment.cached_votes_up == closest_comment_by_votes.cached_votes_up
+          newest_similar_comment = comment.siblings.where(
+            cached_votes_up: closest_comment_by_votes.cached_votes_up).order(updated_at: :asc).last
+          newest_similar_comment.append_sibling(comment)
```
briankung: Woops, I was trying to push to a new branch to submit a PR and git wasn't having it.

```
briankung:projectcredo (votesForComments)$ gp -u origin appendCommentSort
error: src refspec appendCommentSort does not match any.
error: failed to push some refs to 'git@github.com:briankung/projectcredo.git'
Counting objects: 12, done.
Delta compression using up to 4 threads.
Compressing objects: 100% (10/10), done.
Writing objects: 100% (12/12), 1.12 KiB | 0 bytes/s, done.
Total 12 (delta 8), reused 0 (delta 0)
remote: Resolving deltas: 100% (8/8), completed with 4 local objects.
To github.com:briankung/projectcredo.git
   ff4fd3b..c178d76  votesForComments -> votesForComments
Branch votesForComments set up to track remote branch votesForComments from origin.
```

briankung (approved):

app/views/papers/_index.html.erb
wsf246: this is what we were doing before #114

briankung:
Yeah, that's probably better. Especially now that we can count on the
reference being there. Quick note, you can use inject(&:+) instead of
giving it a starting value of 0.

I'm on my phone, but it'd probably be like this:

`comments_count = reference.comments.map {|c|
c.self_and_descendants.length}.inject(&:+)`

Should do the trick.

PR #97, opened 2016-10-03
Overhaul paper views

briankung:
**Problem:** Ugly views, technically speaking

**Solution:** De-uglify them

**Complications:** ~~Still wanna break comments out into their own route~~

I think I'll eventually want to move more of the list functions into a username namespace. So all routes to show any list would look like: `www.projectcredo.com/ws246/allergies-in-immigrant-families`. I also definitely want `/ws246/allergies-in-immigrant-families/:reference_id (of_some_kind, pubmed or a hashid would look fine for instance)` to be a thing.

I _think_ only `list#new` and `list#create` should be at the root level. It starts to get messy when considering `references#show`, `#create`, and `#destroy` and `votes#create` and `#destroy`. But I do think they should be nested on top of `/ws246/allergies-in-immigrant-families/`. For instance: `/ws246/allergies-in-immigrant-families/vote` and `/ws246/allergies-in-immigrant-families/824hdk/vote`for the voting endpoints and `POST` or `DELETE` to `/ws246/allergies-in-immigrant-families/references` to add or remove references.

I'll do the above when I implement friendly IDs.

**Changelog:** 
- Refactored paper/reference views
- Added a reference show route and page
- Tweaked the display of the paper title and like toggle button on reference#show, as well as displaying comments by default
- Created a user profile page to show the lists they've created, linked each commenter's username to it


briankung:
Sorry, this pull request got way out of hand.

app/views/papers/edit.html.erb
```diff
@@ -3,4 +3,3 @@
 <%= render 'form', paper: @paper, button_name: "Update Paper" %>
 
 <%= link_to 'Show', @paper %> |
```
wsf246, line 5: I can still access this page, should we get rid of it temporarily in the controller?  I don't mind having it so I can just edit things if need be without going through console.

PR #99, opened 2016-10-04
Access and import CrossRef data

wsf246:
**Problem:** 
Missing DOI coverage for non-pubmed articles

**Solution:** 
integrate CrossRef API and import using find paper based on locators

**Complications:**


app/models/services/locators/doi_paper_locator.rb
```diff
@@ -6,29 +6,25 @@ def column
...
-        abstract: pubmed.get_abstract(data['uid']),
-        doi: self.locator_id,
-        publication: data['source']
-      )
+    if paper.save
       return paper
     else
-      paper = Paper.new
-      paper.errors.add(:locator_id, "is invalid; no paper found for searched DOI: #{self.locator_id}")
+      paper.errors.add(:locator_id, "is invalid; no paper found for searched DOI: #{self.locator_id}") if paper.errors.empty?
```
briankung: Are you sure you want to add this error if your paper's errors hash is empty?

wsf246: this is prob a short term fix.  there's a bigger issue going on now that we're pulling from multiple sources which is finding papers in credo based on multiple locators.  So the base locator only searches for a paper in our db based on the locator's type.  It will need to search based on doi, uid, or title to remove duplication.  

The reason I put that condition there was because it was throwing a "title is already taken" error if I tried to say submit a DOI for a paper that was already submitted by link and title and didn't have a DOI.  This would trigger the error message "no paper found with this locator id" along side of "Title has already been taken", but it's actually just the second one that was the issue. 

I think if I solve the bigger problem I won't need this conditional.

app/models/services/crossref.rb
```diff
@@ -0,0 +1,24 @@
...
+  private
+
+    def valid_json?(json)
+      begin
+        JSON.parse(json)
+        return true
+      rescue JSON::ParserError => e
+        return false
+      end
+    end
```
briankung, line 24: Normally I'm not much for focusing on efficiency, but you are parsing the JSON 3 times. Also think about what the return value for the method should be. Right now if the condition fails, what's the return value? Whatever called the method will get some other kind of return value that may break things down the line.

wsf246:
yea, I was thinking the same.  The issue I ran into was JSON.parse would
immediately break.  The Net::HTTP would throw an error page in non-JSON.  I
could make the conditional based on that error page, but it seems very
fragile.  Thoughts?  If you have a better method, I'd definitely prefer it.

briankung:
Why not just catch the error in the method? Just make sure to return
something reasonable. It might be better to have the application break
there than to allow a silent failure and have it break somewhere else
unexpected down the line.

app/models/services/crossref.rb
wsf246: You mean run the rescue there?  I got valid from a stackoverflow response so I only kinda know what it does, fault of my ignorance.

Something like:

```
JSON.parse Net::HTTP.get(metadata_uri)
rescue JSON::ParserError => e
```

briankung:
Please don't copy StackOverflow code without understanding every line. Then
please read up on rescuing from errors in ruby.

wsf246:
yes understood.  a bit of a dick-ish response.  I'm trying to get a better understanding of what you're proposing.  My ruby is not as good as yours.

briankung:
Sorry. Rescue blocks start with a begin block where you put dangerous code.
Then you rescue from specific errors with rescue. Finally end the block
with end.

app/controllers/references_controller.rb
```diff
@@ -11,7 +11,7 @@ def create
         flash['notice'] = 'This paper has already been added to this list'
       else
         Reference.create(list_id: list.id, paper_id: paper.id)
-        flash['notice'] = "You added #{paper.title} to #{list.name}"
+        flash['notice'] = "You added '#{paper.title}'' to '#{list.name}'"
```
briankung: I think I have a similar change in my WIP branch. It's fine - I'll manage the merge when the conflict happens. However, you are erroneously using two apostrophes. at the end of `#{paper.title}`

app/models/services/crossref.rb
```diff
@@ -0,0 +1,17 @@
...
+    @metadata_url = @base_url + "works/"
+    @funder_url = @base_url + "funders/"
+  end
+
+  def get_doi_metadata(doi)
+    metadata_uri = URI.parse(@metadata_url+doi)
+    begin
+      JSON.parse Net::HTTP.get(metadata_uri)
+    rescue JSON::ParserError => e
+      return false
```
briankung: What does returning `false` in the context of `get_doi_metadata` mean, though? A common convention in Ruby is that when a return value is supposed to be an object, it either raises an error or returns `nil`. This way you can either crash quickly or return an object that means, essentially, the object in question wasn't found. That's how the ActiveRecord `#find_by` method works.

app/models/services/locators/base_locator.rb
```diff
@@ -8,4 +8,50 @@ def initialize locator_id
...
+
+      paper.title ||= imported_data['title'].first
+      paper.published_at ||=  imported_data['created']['date-time'].to_date
+      paper.doi ||= imported_data['DOI']
+      paper.publication ||= imported_data['publisher']
+      paper.link ||= imported_data['link'].first['URL'] if imported_data['link']
+    end
+
+    return paper
+  end
```
briankung: I'm not comfortable with this solution. Besides which, it almost certainly should not be in the base class. We definitely need to create different mappings per service, but those mappings should not be here.

app/models/services/locators/doi_paper_locator.rb
```diff
@@ -6,30 +6,32 @@ def column
   def find_paper
     return super if super
 
+    paper = Paper.new
+
+    crossref = Crossref.new
+    if result = crossref.get_doi_metadata(self.locator_id)
```
briankung: Just a small note, I've been making it clear that a variable has been assigned in the condition by putting the conditional assignment in parentheses. So it would look like `(result = crossref.get_doi_metadata(self.locator_id))`. Not a big deal, but a stylistic preference.

app/models/services/crossref.rb
briankung: Another convention to keep in mind is that it makes more sense if certain methods have certain return values. For instance, a method that returns `false` should also return `true`. In the case that it doesn't return a boolean, then perhaps its negative value should be `nil`.

app/models/services/locators/doi_paper_locator.rb
```diff
@@ -6,30 +6,32 @@ def column
...
       result = pubmed.get_uid_metadata(uid)
       data = result['result'][uid]
-      names = data['authors'].map {|a| a['name']}
+      paper = self.import_data_to_paper(paper,data,'pubmed')
+    end
 
-      existing_authors = Author.where(name: names)
-      existing_names = existing_authors.map(&:name)
-      new_authors = (names - existing_names).map {|a| Author.create name: a}
+    existing_paper = Paper.where( "title = ? or pubmed_id = ? and pubmed_id is not null", paper.title, paper.pubmed_id )
```
briankung: I would prefer the Rails finders here. `Paper.where(title: paper.title, pubmed_id: paper.pubmed_id).not(pubmed_id: nil)`, although something about that strikes me as odd, anyway. If you're using the paper's pubmed id, why do you have to check if the pubmed id is null?

app/models/services/locators/pubmed_paper_locator.rb
```diff
@@ -6,35 +6,31 @@ def column
...
+    if paper.doi
+      crossref = Crossref.new
+      if result = crossref.get_doi_metadata(self.locator_id)
+        data = result['message']
+        paper = self.import_data_to_paper(paper,data,'crossref')
       end
+    end
 
-      doi = doi.sub(/^doi: /, "")
+    existing_paper = Paper.where( "title = ? or doi = ? and doi is not null", paper.title, paper.doi )
```
briankung: Same here, I would prefer `Paper.where(title: paper.title, doi: paper.doi).not(doi: null)`. Similarly, I wonder why you need to check if the doi is null.

briankung: `#where` also returns a query - which means semantically that it can return multiple results. Essentially it returns an array that can even be empty.

app/models/services/locators/pubmed_paper_locator.rb
```diff
@@ -6,35 +6,31 @@ def column
...
-        pubmed_id: data['uid'],
-        title: data['title'],
-        published_at: data['pubdate'],
-        authors: (existing_authors + new_authors),
-        abstract: pubmed.get_abstract(data['uid']),
-        doi: doi,
-        publication: data['source']
-      )
+    if existing_paper.present?
+      return existing_paper
```
briankung, line 29: Same note on `#where` returning the equivalent of an array

app/models/services/locators/doi_paper_locator.rb
```diff
@@ -6,30 +6,32 @@ def column
...
-        pubmed_id: data['uid'],
-        title: data['title'],
-        published_at: data['pubdate'],
-        authors: (existing_authors + new_authors),
-        abstract: pubmed.get_abstract(data['uid']),
-        doi: self.locator_id,
-        publication: data['source']
-      )
+    if existing_paper.present?
+      return existing_paper
```
briankung, line 28: See note below on `#where` returning an array

app/models/services/locators/pubmed_paper_locator.rb
```diff
@@ -6,35 +6,31 @@ def column
...
+    elsif paper.save
       return paper
     else
-      paper = Paper.new
-      paper.errors.add(:locator_id, "is invalid; no paper found for searched Pubmed ID: #{self.locator_id}")
+      paper.errors.delete(:title)
```
briankung, line 33: I forgot why this is necessary, again?

briankung (changes requested):
I will probably step in to do some architecting with the mappings. There are some code smells, but otherwise it looks fine for the most part.

We do have some thinking to do about reconciling imported paper data from multiple sources.

app/models/services/locators/doi_paper_locator.rb
wsf246: This is for papers that have DOIs but no pubmed_ids and vice versa.  If I import a paper from crossref that doesn't have a pubmed_id, then the #where it will return all the papers where pubmed_id is null as well.

app/models/services/locators/doi_paper_locator.rb
wsf246: yea, I chose where because I couldn't find an easy way to do an "or" statement for 2 attributes using #find.  It was easier to do the "or" in the where statement to broaden and eliminate duplication.  

The "or" is for the case where they may have entered a paper title without a pubmed id or doi, or if the title was imported slightly differently. I'll test your suggestion to see if I get the same result.

app/models/services/locators/pubmed_paper_locator.rb
wsf246: I should probably rewrite this to make it more clear.  The current mapping creates a new Paper and imports crossref and Pubmed data if it can locate it in their api's.  If it can't find it in either, it will return a paper with a blank title which won't save.  However, this will also return a "title can't be blank" error, which I'm getting rid of.

I can probably add something to the "else" to show the title is nil because it didn't find anything in pubmed or crossref.  That would make it more readable.

app/models/services/locators/base_locator.rb
wsf246: Ok, I'll move it into each locator.  Should I separate the crossref and pubmed stuff?  Should i even have it as a separate method?

briankung: Probably keep the same name. Although I might do that entirely differently. I would move it into a locator, but I would have a mappings class that we could hot swap out for different services. Not sure what I would want yet.

app/models/services/locators/doi_paper_locator.rb
briankung: Oh, I see. I missed the `or` in there. Carry on!

app/models/services/locators/doi_paper_locator.rb
briankung: You probably won't, I didn't see the "or." But if you use a `#where` statement, you should probably pick one of (oftentimes the `#first`) results.

app/models/services/locators/pubmed_paper_locator.rb
briankung: That makes sense. We are probably doing something wrong with this strategy if we have to do something janky like that, but it's probably fine for now.

app/models/services/locators/base_locator.rb
briankung: Anyway, don't worry about it so far as the hot swappable map classes. Let's just go with methods of the same name on each locator for now.

wsf246:
That should be all the changes, let me know if it looks good to merge.  I also added an additional piece to grab an abstract directly from the pubmed site if it doesn't exist in the eutils db, but it's a little janky.  The ones I couldn't find a few days ago are now up.

Another item for consideration for me is the potential for different papers to have the same title.  Probably something we can deal with later, but I've seen cases where there will be something like "Effects of aspartame on cancer" or something that shows up multiple times.

briankung:
Hmm so basically we shouldn't even have that validation.

wsf246:
yea, prob in the long run.  Although if we can validate on title + first
author, i think that would work for just about everything.

app/models/paper.rb
```diff
@@ -9,7 +9,7 @@ class Paper < ApplicationRecord
 
   accepts_nested_attributes_for :authors, reject_if: proc { |attributes| attributes['name'].blank? }
   validates_associated :authors
-  validates :title, uniqueness: true
+  validates :title, uniqueness: true, presence: true
```
briankung, line 12: I just added issue #101 to address uniqueness by title + something else

app/models/services/crossref.rb
```diff
@@ -0,0 +1,17 @@
...
+    @funder_url = @base_url + "funders/"
+  end
+
+  def get_doi_metadata(doi)
+    metadata_uri = URI.parse(@metadata_url+doi)
+    begin
+      JSON.parse Net::HTTP.get(metadata_uri)
+    rescue JSON::ParserError => e
+      return nil
+    end
```
briankung, line 14: Let's just let this break if it fails. Otherwise your other methods will keep going forward with `nil` values and who knows what will happen.

briankung: I mean, let's remove the rescue block.

app/models/services/locators/doi_paper_locator.rb
```diff
@@ -6,30 +6,79 @@ def column
   def find_paper
     return super if super
 
+    paper = Paper.new
+
+    crossref = Crossref.new
+    if (result = crossref.get_doi_metadata(self.locator_id))
```
briankung: Let's get rid of these if statements so that the whole process just explodes if the `get_doi_metadata` method fails.

app/models/services/locators/doi_paper_locator.rb
```diff
@@ -6,30 +6,79 @@ def column
...
+      data = result['message']
+      paper = self.import_data_to_paper(paper,data,'crossref')
+    end
+
     pubmed = Pubmed.new
-    if uid = pubmed.find_uid_by_doi(self.locator_id)
+    if (uid = pubmed.find_uid_by_doi(self.locator_id))
```
briankung, line 19: Ditto with this if statement.

app/models/services/locators/doi_paper_locator.rb
```diff
@@ -6,30 +6,79 @@ def column
...
+        new_authors = (names - existing_names).map {|a| Author.create name: a}
+        paper.authors = (existing_authors + new_authors)
+      end
+
+      paper.title ||= imported_data['title'].first
+      paper.published_at ||=  imported_data['created']['date-time'].to_date
+      paper.doi ||= imported_data['DOI']
+      paper.publication ||= imported_data['publisher']
+      paper.link ||= imported_data['link'].first['URL'] if imported_data['link']
+    end
```
briankung: All of this code looks like it's duplicated in both the doi locator and the pubmed locator. Is this necessary? Also, it's a gigantic method, regardless.

app/models/services/pubmed.rb
```diff
@@ -4,6 +4,8 @@ def initialize(options={})
     @search_url = @base_url + "/entrez/eutils/esearch.fcgi"
     @metadata_url = @base_url + "/entrez/eutils/esummary.fcgi"
     @abstract_url = @base_url + "/entrez/eutils/efetch.fcgi"
+    @pubmed_url = 'https://www.ncbi.nlm.nih.gov/pubmed/'
+    @pubmed_xml_options = '?report=xml&format=text'
```
briankung: Instead of this, try using the `generate_uri` method. If you put these key/value pairs into the hash at the end of method it will add the parameters. Look at other uses of `#generate_uri` in this file.

app/models/services/pubmed.rb
```diff
@@ -34,10 +36,15 @@ def get_uid_metadata(uid)
   def get_abstract(uid)
     abstract_uri = generate_uri(@abstract_url, @default_parameters.merge(id: uid, retmode: 'xml'))
 
-    abstract = Hash.from_xml(Net::HTTP.get(abstract_uri))['PubmedArticleSet']['PubmedArticle']['MedlineCitation']['Article']['Abstract']['AbstractText']
+    begin
+      abstract = Hash.from_xml(Net::HTTP.get(abstract_uri))['PubmedArticleSet']['PubmedArticle']['MedlineCitation']['Article']['Abstract']['AbstractText']
+    rescue
+      pubmed_scrape = Hash.from_xml(Net::HTTP.get(URI.parse(@pubmed_url + uid + @pubmed_xml_options)))
+      abstract = pubmed_scrape['pre'][/#{"AbstractText"}(.*?)#{"</AbstractText>"}/m, 1].partition('>').last
```
briankung: Please break these into separate lines. That will help you say where each one broke if there's an error and it's much more readable.

briankung: Also, you seem to have fragile code in the rescue block, as well. If that code breaks, what happens then?

briankung (changes requested):

app/models/services/pubmed.rb
briankung: Additionally, I just found out about Ruby's `#dig` method: http://ruby-doc.org/core-2.3.1/Hash.html#method-i-dig

Seems pretty cool

app/models/services/crossref.rb
wsf246: The rescue block handles doi's that aren't found in crossref.  [Basically the api returns this if it can't find the doi:](http://api.crossref.org/works/10.1037/poop)

For this method, it's meant to grab the data if it exists.  The JSON parse breaks if the doi is not found because the API doesn't return JSON.  So I have this method return a nil and  I handle that nil in each #find_paper for the locators:

```
    crossref = Crossref.new
    if (result = crossref.get_doi_metadata(self.locator_id))
      data = result['message']
      paper = self.import_data_to_paper(paper,data,'crossref')
    end
```

The reason I have it continue is so I can search pubmed for that doi as well.  Crossref's API only has doi's listed by crossref, which should be most, but there's a chance that the doi will exist in pubmed and not crossref.

Alternatively, I can handle the JSON parse outside this method if we want to remove this error block from this method.  I was trying to follow the convention of "return an object or nil" but let me know if there's a better way.

app/models/services/locators/doi_paper_locator.rb
wsf246: same kind of idea as above.  The nil process is handling the "doi not in crossref" case.  I could do something more specific to the "resource not found" case.

app/models/services/locators/doi_paper_locator.rb
wsf246: So this is the same issue.  I guess it ends up being a UI question.  Essentially these if statements are to keep the user from hitting an error page if they enter pubmed ID or DOI that is not in the APIs, either it's not covered or they typed it in wrong.  The error messages are handled in the references controller based on what is spit out.  

It seems to make more sense to have an alert pop up and have them sent back to the list show page instead of a broken page?

app/models/services/locators/doi_paper_locator.rb
wsf246: yea, this was the method that I pulled out from base locator that we wanted in each locator.  This was the one you thought about doing a hot swap with?

app/models/services/pubmed.rb
wsf246: I wrote stuff but accidentally submitted it as a review, then cancelled the review and it's gone.  Did it get sent via email?

app/models/services/crossref.rb
briankung: This isn't really an object - in this case, it's a return value from an HTTP request. On our side, it's just data, or a series of attributes. We're not retrieving a record or interacting with some object. In this case, to handle the error of a bad DOI like http://api.crossref.org/works/10.1037/poop, we can just check the HTTP status code. If you open up the network tab while loading that URL, you can see that it's a 404. We can check to see if an HTTP request returns a success code (2xx range) or perhaps a redirect and then a 200 if we're getting fancy, and then handle the result accordingly.

Also, we should use rescue blocks sparingly. It's not really an application level breakage if an HTTP request returns with a non-success status code. https://webcache.googleusercontent.com/search?q=cache:9LlMiCOwyOkJ:https://httpstatusdogs.com/+&cd=1&hl=en&ct=clnk&gl=us

briankung: Essentially you'll need to break the Net::HTTP.get request out and check it before shoving it into a JSON parser.

app/models/services/locators/doi_paper_locator.rb
briankung: If it really actually catastrophically fails, then we should let it fail. Otherwise, we should handle 4xx and 5xx responses from the external API gracefully: https://github.com/briankung/projectcredo/pull/99/files/421f3270169e3dadbce1f99dff2edfbee31b426c#r82536967

app/models/services/locators/doi_paper_locator.rb
briankung: It does, but let's use the proper level of error handling. https://github.com/briankung/projectcredo/pull/99/files/421f3270169e3dadbce1f99dff2edfbee31b426c#r82536967

app/models/services/locators/doi_paper_locator.rb
briankung: "Hot swappable class" in that instance meant something like the locators, which have a common interface / API, so you can replace it with a different class that implements the same methods. Also known as duck typing.

I was talking about having a `PubmedArticleApiPaperMap` or a `CrossRefApiPaperMap` that could take the attributes that each respective API returned and map them to their respective attributes in our `Paper` model.

These methods are gigantic, which is a code smell, and they _seem_ to duplicate logic that should be particular to Pubmed (`if source == "pubmed"`) or Crossref (`if source == "crossref"`) respectively. Do you need to handle both sources in both places? If so, is there any refactoring you can do?

app/models/services/pubmed.rb
briankung: Unfortunately, it did not.

briankung:
As long as you test this and resolve that merge conflict, this is good to go 👍

briankung (approved):

briankung:
\o/

PR #100, opened 2016-10-08
Username validation

briankung:
**Problem:** I was just checking the username validation and found some possible improvements

**Solution:** Tightened up the regex, added a more informative error message

**Complications:** None

**Changelog:** 
- Stops allowing multiline usernames
- Required 3+ characters `[a-zA-Z_]`
- Added a more informative error message

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -8,7 +8,7 @@ class User < ApplicationRecord
   # This is in addition to a real persisted field like 'username'
   attr_accessor :login
 
-  validates_format_of :username, with: /^[a-zA-Z0-9_\.]*$/, multiline: true
+  validates_format_of :username, with: /\A[a-zA-Z0-9_\.]*\z/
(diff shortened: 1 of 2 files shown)


app/models/user.rb
```diff
@@ -8,7 +8,7 @@ class User < ApplicationRecord
   # This is in addition to a real persisted field like 'username'
   attr_accessor :login
 
-  validates_format_of :username, with: /^[a-zA-Z0-9_\.]*$/, multiline: true
+  validates_format_of :username, with: /\A[a-zA-Z0-9_\.]{3,}\z/
```
briankung, line 11: I just realized that this doesn't allow for unicode characters like 世界. I think we can do better.

wsf246: should I wait to merge or open up a new issue?

briankung:
You can merge if you want

PR #112, opened 2016-10-11
Fix schema

### LINKED ISSUE #110: Add missing indices to speed up lookups

### LINKED ISSUE #111: Convert id fields to integers

wsf246:
**Problem:** 
No indexes on foreign keys that cause performance hits with large DB.  Some IDs should be strings
Closes #111 
Closes #110 
**Solution:** 
Add indexes where needed
**Complications:** 

**Changelog:** 

Add indexes to all foreign keys
Cleanup some sort indexes
Convert string IDs to integer


wsf246:
This should cover all the foreign key cases, let me know if I missed any and your thoughts on whether to consolidate some of the indexes into a single one.  I built these indexes based on how I think we will be searching the DB through associations, direct lookups, or sorts.

briankung:
lgtm 👍

PR #119, opened 2016-10-16
Multiple links

### LINKED ISSUE #102: Paper link should validate properly
This was helpful: http://stackoverflow.com/a/7168125/1042144

I added it, intending to submit a quick PR, but then discovered that something is wrong with the `persisted?` check in the references controller. Since this is changing in #99, I thought I'd just make an issue.

### LINKED ISSUE #109: Support multiple Links in Papers

wsf246:
**Problem:** 
Closes #109
Closes #102

**Solution:** 
Create Links table, associate, and validate links. 

**Complications:** 
I'm going to add the ability to add links to each paper when I adjust the form for each paper.

**Changelog:** 
- Create Links table
- Associate with paper has many relationship
- Grab multiple links from crossref
- Validate links
- Fix error messages


app/models/link.rb
```diff
@@ -0,0 +1,10 @@
+class Link < ApplicationRecord
+  belongs_to :paper
+  validates_presence_of :url
+  validate :valid_url
```
briankung: I'm pretty sure you can do both of the validations with one call to validates. Also, did you try using the regex I mentioned in #102?

briankung: The reason I'm curious is because I think it translates it to a frontend regex which prevents the form from even being sent to the backend if it doesn't validate. It's a bit more efficient.

Either that, or I'm misremembering and that's only if you're using a particular gem.

PR #120, opened 2016-10-17
Routing via slugs

### LINKED ISSUE #108: Use Friendly IDs in routes

### LINKED ISSUE #115: Revamp routes
I think I'll eventually want to move more of the list functions into a username namespace. So all routes to show any list would look like: `www.projectcredo.com/ws246/allergies-in-immigrant-families`. I also definitely want `/ws246/allergies-in-immigrant-families/:reference_id (of_some_kind, pubmed or a hashid would look fine for instance)` to be a thing.

I _think_ only `list#new` and `list#create` should be at the root level. It starts to get messy when considering `references#show`, `#create`, and `#destroy` and `votes#create` and `#destroy`. But I do think they should be nested on top of `/ws246/allergies-in-immigrant-families/`. For instance: `/ws246/allergies-in-immigrant-families/vote` and `/ws246/allergies-in-immigrant-families/824hdk/vote`for the voting endpoints and `POST` or `DELETE` to `/ws246/allergies-in-immigrant-families/references` to add or remove references.

briankung:
**Problem:** The routes we have right now are not terribly friendly. It would be nice to support `projectcredo.com/briankung/allergies-in-immigrant-families/124`. Closes #115, closes #108 

**Solution:** I've added some of the routes from #115, but I've left a lot of top level routes.

**Complications:** The problem with even moderate nesting is that we have to maintain knowledge about the chain of parent objects forever and ever and I haven't thought of a good way to do that.

**Changelog:** 
- Add slugs to lists and logic for converting to slugs
- Removes a bunch of unused code
- Undo and redo routes forever until I found a simple solution
- Adds user-namespaced views/controllers
- Use explicit render calls in views and controllers for portability
- Unbreaks everything I can think of after changing routes


app/controllers/lists_controller.rb
```diff
@@ -68,7 +68,7 @@ def destroy
   private
     # Use callbacks to share common setup or constraints between actions.
     def set_list
-      @list = List.find(params[:id])
+      @list = List.find_by(slug: params[:id])
```
wsf246, line 71: :+1:

app/controllers/papers_controller.rb
```diff
@@ -78,15 +37,4 @@ def paper_params
...
-    def params_sort_order
-      pub_date_order = 'papers.published_at DESC NULLS LAST'
-      vote_order = 'total_votes DESC'
-
-      if params[:sort] == 'pub_date'
-        pub_date_order
-      else # default to ordering by votes first, then pub date
-        "#{vote_order}, #{pub_date_order}"
-      end
-    end
```
wsf246, line 91: hazaa for removal of unused code

PR #122, opened 2016-10-18
user goes to last shown page on successful sign-up/sign-in

wsf246:
**Problem:** 
Signin/Signup annoyingly sends back to root
**Solution:** 
Use Devise solution to send signins/signup to last shown page
https://github.com/plataformatec/devise/wiki/How-to:-redirect-to-a-specific-page-on-successful-sign-in-out


briankung:
Man, these Devise wiki pages can get really janky.

PR #128, opened 2016-10-21
Ajax comments

### LINKED ISSUE #92: User should not reload the entire page when voting or commenting
use JS to stop reload of page after every vote or new comment

wsf246:
**Problem:** 
adds ajax to comments
**Solution:** 
uses ujs to create update edit and destroy comments. Closes #92 
**Complications:** 
I don't think the comments controller is validating the current user correctly
**Changelog:** 
-adds ujs for create, edit, update, and destroy
-removes options for non-comment creator
-updates comments count


app/controllers/comments_controller.rb
```diff
@@ -11,14 +11,16 @@ def edit
   def create
     @comment = Comment.new(comment_params)
     @comment.user = current_user
-
     respond_to do |format|
       if @comment.save
+        @reference = Reference.find(@comment.root.commentable_id)
         format.html { redirect_to :back, notice: 'Comment was successfully created.' }
         format.json { render :show, status: :created, location: @comment }
+        format.js { render 'cud.js.erb' }
```
briankung: Please rename this partial.

app/controllers/comments_controller.rb
```diff
@@ -28,22 +30,27 @@ def create
   def update
     respond_to do |format|
       if @comment.update(comment_params)
+        @reference = Reference.find(@comment.root.commentable_id)
```
briankung: If you're using `commentable_id`, you can just do `@reference = @comment.root.commentable`

app/views/comments/cud.js.erb
```diff
@@ -0,0 +1,6 @@
+$("#comments-<%= @reference.id %>").html("<%= escape_javascript(render 'papers/comments', reference: @reference) %>");
```
briankung: Instead of a class variable, let's have the partial take it as a local variable. Makes it easier to reuse later - you don't need to set up the controller to have a `@reference` variable, you just feed a reference into the partial.

app/views/comments/cud.js.erb
```diff
@@ -0,0 +1,6 @@
...
+$("li#comments-tab-<%= @reference.id %>").html("<%= escape_javascript(render 'papers/comments_tab', reference: @reference) %>");
+$('.toggle-reply').on('click', function(e) {
+  e.preventDefault();
+  $(e.target).parents("ul").next('.comment-reply').toggleClass('hidden');
```
briankung: We should probably find a better way to do this

app/views/papers/_comments_tab.html.erb
```diff
@@ -0,0 +1,5 @@
+ <%
+  comments_count = reference.comments.map {|c| c.self_and_descendants.length}.inject(0,&:+)
+  display_count = comments_count.zero? ? "" : "(#{comments_count})"
+%>
+<%= link_to "Comments#{display_count}", '', "data-detail-id" => "comments-#{reference.id}", class: " paper-detail-tab" %>
```
briankung: 👍  to breaking this into a partial

app/views/comments/edit.js.erb
```diff
@@ -0,0 +1,11 @@
...
+  .html(
+    "<%= escape_javascript(render 'form', comment: @comment, display_comment_field: true )%>"
+  )
+  .find(".btn").after( "<small>   <%= escape_javascript(link_to 'Cancel', '#') %></small>" )
+  .next().find("a").on('click', function(e) {
+    e.preventDefault();
+    $(e.target).parents("#comment-content<%= @comment.id %>").html( "<%= escape_javascript(@comment.content) %>" );
+  });
+;
+
```
briankung, line 3: This file is extremely difficult to read. Is there anything you can do to increase the readability of what you're doing here?

briankung (changes requested):

app/controllers/comments_controller.rb
wsf246: done

app/controllers/comments_controller.rb
wsf246: done

app/views/comments/cud.js.erb
wsf246: done

app/views/comments/cud.js.erb
wsf246: I agree.  You mean specifically this function or rewriting this in this render.  I'm not 100% sure why me rendering papers/comment_tab again stops the reply button from working, maybe because the js we have in comments.js only starts after the DOM has loaded?

app/views/comments/edit.js.erb
wsf246: done

wsf246:
I think that should be all the points.  Let me know what you think.

app/views/comments/reference_comments.js.erb
```diff
@@ -0,0 +1,6 @@
+$("#comments-<%= reference.id %>").html("<%= escape_javascript(render 'papers/comments', reference: reference) %>");
+$("li#comments-tab-<%= reference.id %>").html("<%= escape_javascript(render 'papers/comments_tab', reference: reference) %>");
+$('.toggle-reply').on('click', function(e) {
+  e.preventDefault();
+  $(e.target).parents("ul").next('.comment-reply').toggleClass('hidden');
```
briankung: We should probably just give each comment reply its own id and select based on that somehow. We can do it later, though.

app/views/comments/edit.js.erb
```diff
@@ -0,0 +1,13 @@
...
+commentSubmitButton = commentContentP.find(".btn")
+commentSubmitButton.after("<small>   <%= escape_javascript(link_to 'Cancel', '#') %></small>")
+
+cancelLink = commentSubmitButton.next().find("a");
+cancelLink.on('click', function(e) {
+  e.preventDefault();
+  commentContentP.html("<%= escape_javascript(@comment.content) %>");
+});
+
+
```
briankung: This file is waayyyy more readable \o/

briankung (approved):
This looks good to me. You can make that minor change I suggested or not, up to you 👍

briankung:
Just be aware of the merge conflicts.

app/controllers/comments_controller.rb
```diff
@@ -11,14 +11,16 @@ def edit
   def create
     @comment = Comment.new(comment_params)
     @comment.user = current_user
-
     respond_to do |format|
       if @comment.save
+        reference = @comment.root.commentable
         format.html { redirect_to :back, notice: 'Comment was successfully created.' }
         format.json { render :show, status: :created, location: @comment }
+        format.js { render( @comment.parent.nil? ? 'comment_on_reference.js.erb' : 'reply_to_comment.js.erb', locals: {reference: reference} )}
```
wsf246: so this is the part I was struggling with yesterday, but I've got a solution and wanted your opinion on it.  Basically because commenting on a reference is different enough than replying to a comment, I seperated out the logic into 2 different jquery pieces.  There's some duplication, so I was considering joining these two things with an if statement in the jquery, but this seems cleaner and I feel like it makes more sense to make the controller do that.  Thoughts?

app/views/comments/reply_to_comment.js.erb
```diff
@@ -0,0 +1,14 @@
+commentParentDiv = $("#comment-<%= @comment.parent_id %>");
+replyDiv = commentParentDiv.children('.reply');
```
wsf246: the main difference between replying to a comment or a reference is the reply div wrapper and seeing if it exists already.

app/assets/javascripts/comments.js
```diff
@@ -1,6 +1,7 @@
 $(document).ready(function() {
-  $('.toggle-reply').on('click', function(e) {
+  $(document).on('click', '.toggle-reply', function(e) {
     e.preventDefault();
-    $(e.target).parents("ul").next('.comment-reply').toggleClass('hidden');
+    commentID = $(e.target).data('parent-comment-id');
+    $(e.target).parents("ul").next("#reply-form-"+commentID).toggleClass('hidden');
```
briankung: Since we're using an this id `#reply-form-"+commentID` we don't have to do the `$().parents().next()` thing. Since there should be only one - hence an ID - you can just grab the ID: `$("#reply-form-"+commentID)`

app/controllers/comments_controller.rb
```diff
@@ -28,22 +30,27 @@ def create
...
         format.json { render json: @comment.errors, status: :unprocessable_entity }
+        format.js { render 'reference_comments.js.erb', locals: {reference: reference} }
       end
     end
   end
 
   # DELETE /comments/1
   # DELETE /comments/1.json
   def destroy
+    reference = @comment.root.commentable
```
briankung, line 53: I forgot to note earlier that this will stop working if we add another model as commentable

briankung: `¯\_(ツ)_/¯`

app/views/comments/reply_to_comment.js.erb
```diff
@@ -0,0 +1,14 @@
...
+
+if (replyDiv.length) {
+  replyDiv.children().last().after("<%= escape_javascript(render('comments/comment', comment: @comment)) %>");
```
briankung: I think you can just do `replyDiv.append()`

app/views/comments/reply_to_comment.js.erb
```diff
@@ -0,0 +1,14 @@
...
+} else {
+  commentParentDiv.children().last().after(
+    "<%= escape_javascript(
+      content_tag( :div, render('comments/comment', comment: @comment), class: 'reply')
+    )%>"
+  );
```
briankung: Same here with `append()`

app/views/comments/reply_to_comment.js.erb
briankung: I feel like this should not be the case...let's see what we can do instead.

app/controllers/comments_controller.rb
briankung: I feel like it should only require one jquery partial.

Also, you and your ternary operators. When these ternary operator lines get really long, I recommend breaking them out into `if/else` statements.

For this one in particular, you can do 

``` ruby
format.js do
  if @comment.parent
    render 'reply_to_comment.js.erb', locals: {reference: reference}
  else
    render 'comment_on_reference.js.erb', locals: {reference: reference}
  end
end
```

Way more readable.

That said, I strongly feel like it shouldn't require two separate partials.

app/views/comments/destroy.js.erb
```diff
@@ -0,0 +1,5 @@
+commentDiv = $("#comments-<%= reference.id %>").find("#comment-<%= @comment.id %>");
+if (commentDiv.siblings().length == 0) { commentDiv.parent(".reply").remove() };
+commentDiv.remove();
+
+$("li#comments-tab-<%= reference.id %>").html("<%= escape_javascript(render 'references/comments_tab', reference: reference) %>");
```
briankung: Would you mind just decrement the counter instead of re-rendering the entire comments tab? You'd probably want to use JavaScript `parseInt`

briankung: Then subtract one and reinsert it

app/views/comments/edit.js.erb
```diff
@@ -0,0 +1,11 @@
...
+commentContentP.html("<%= escape_javascript(render 'form', comment: @comment, display_comment_field: true )%>");
+
+commentSubmitButton = commentContentP.find(".btn")
+commentSubmitButton.after("<small>   <%= escape_javascript(link_to 'Cancel', '#') %></small>")
+
+cancelLink = commentSubmitButton.next().find("a");
+cancelLink.on('click', function(e) {
+  e.preventDefault();
+  commentContentP.html("<%= escape_javascript(@comment.content) %>");
+});
```
briankung, line 16: Is re-registering the handler necessary, or can you use the `$(document).on('event', 'selector' function)` live handler syntax?

app/controllers/comments_controller.rb
briankung: Let's pair program on this one soon.

briankung (changes requested):

briankung:
Let's pair program on this one soon

wsf246:
yea, let's and finish this guy up.  When is good for you?

app/assets/javascripts/comments.js
wsf246: I do this if I can because it limits the amount of jquerying so it doesn't have to hunt through the whole DOM to find it.  It's minor and not a big deal so I'm fine with whichever way.

app/controllers/comments_controller.rb
wsf246: lol, we'll cross that bridge when we get there.

app/views/comments/edit.js.erb
wsf246: mmm, not sure, let's pair program on this one.

app/assets/javascripts/comments.js
briankung: `$().parents().next()` is much more fragile if the HTML changes, which it probably will. I also think hunting through the DOM for an ID is pretty fast, considering there should be only one of them.

app/controllers/comments_controller.rb
```diff
@@ -11,11 +11,33 @@ def edit
...
     respond_to do |format|
       if @comment.save
+        reference = @comment.root.commentable
+
+        if @comment.parent
+          parent_id = @comment.parent_id
+          new_comment = @comment.parent.children.build
+          display_comment = false
+        else
+          parent_id = "r#{reference.id}"
```
briankung: This doesn't seem like a good idea. I'd prefer to switch based on a class or some way of indicating the type of element you want to anchor the comment to.

app/controllers/comments_controller.rb
```diff
@@ -11,11 +11,33 @@ def edit
...
         format.html { redirect_to :back, notice: 'Comment was successfully created.' }
         format.json { render :show, status: :created, location: @comment }
+        format.js {
+          render(
+            'create.js.erb',
+            locals: {
+              reference: reference,
+              parent_id: parent_id,
+              new_comment: new_comment,
+              display_comment: display_comment
```
briankung: I could go either way on this issue, but this seems like a lot of locals to be cramming into one partial.

app/controllers/comments_controller.rb
```diff
@@ -34,7 +36,8 @@ def create
               reference: reference,
               parent_id: parent_id,
               new_comment: new_comment,
-              display_comment: display_comment
+              display_comment: display_comment,
+              comments_class: comments_class
```
wsf246: I added this class to each of the child comment divs to specify if it was for a reference or a comment.  Let me know if you can think of a more elegant way to do this, maybe putting it in the ID name itself.

wsf246:
it's a lot of local variables to push to that partial, but they're largely due to the way closure tree is setup and how we need to build a comment.  I think it's fine for now and we can refactor or simplify later on, especially if we eventually move to disqus or something, probably more meaningful to push this working set of code for now.

app/views/comments/destroy.js.erb
```diff
@@ -0,0 +1,7 @@
+comments = $("#comment-<%= @comment.id %>").find(".comment").length + 1;
+console.log(comments);
```
briankung: Debug statement

app/controllers/comments_controller.rb
briankung: Let's revert this change, actually. Doesn't look like it serves a purpose.

app/views/comments/destroy.js.erb
wsf246: shit, good catch.

wsf246:
If this looks good, I'll merge.  I can leave some comments in the controller so we can come back to all those local variables getting pushed in later.

briankung (changes requested):
Let's use comment generations and sort order to target a comment or a generation to append a comment.

Basically it works like this:

```
comment A (generation 0)
    comment B (generation 1)
    comment C (generation 1)
        comment D (generation 2)
    comment E (generation 1)
        comment F (generation 2)
```

I'm writing up how to do this at the moment

briankung:
I'm just going to refactor this myself, if you don't mind

wsf246:
yea go for it.

app/views/comments/_comment.html.erb
```diff
@@ -13,7 +13,7 @@
       <li><%= link_to 'delete', comment_path(comment), remote: true, class: "btn btn-link btn-xs", method: :delete %><li>
     <% end %>
   </ul>
-  <%= (render 'comments/form', comment: comment.children.build, display_comment_field: false) if current_user %>
+  <%= render('comments/form', comment: comment.children.build, hidden: true) if current_user %>
```
wsf246, line 16: smart.

app/views/comments/_form.html.erb
```diff
@@ -1,10 +1,12 @@
...
+<% hidden ||= false %>
+
+<%= form_for comment, html: {id: "reply-form-#{comment.parent_id}", class: "comment-reply #{'hidden' if hidden}"}, remote: true do |f| %>
   <div class="form-group">
     <%= f.text_area :content, class: 'form-control comment-box', placeholder: 'Write a reply...' %>
   </div>
   <%= f.hidden_field :parent_id, value: comment.parent_id %>
-  <%= f.hidden_field :commentable_type, value: comment.commentable_type if display_comment_field %>
-  <%= f.hidden_field :commentable_id, value: comment.commentable_id if display_comment_field %>
+  <%= f.hidden_field :commentable_type, value: comment.commentable_type %>
```
wsf246, line 8: good refactor

app/controllers/comments_controller.rb
```diff
@@ -11,11 +11,30 @@ def edit
...
         format.html { redirect_to :back, notice: 'Comment was successfully created.' }
         format.json { render :show, status: :created, location: @comment }
+        format.js do
+          reference = @comment.root.commentable
+
+          if @comment.parent
+            parent_id = @comment.parent_id
+            new_comment = @comment.parent.children.build
+          else
+            parent_id = "r#{reference.id}"
```
wsf246: looks like you reverted back to having the "r" in the class.  Are you changing up this structure to match the generations?

wsf246:
hey, just checking if you're still working on this or if you were looking to merge.  Looks like the list breaks on list show on the heroku pull request.

I was thinking if you were going to go for the generation class or id for the child comments div, you should probably include the reference ID.  It might be something like child-comments-ref-<%= ref.id %>-comment-gen-<%= gen %>.  The generation may not be granular enough either because gen 2 can sit under different comments depending on how you implement and if you are tracking based on the comment ID as well.

briankung:
I'm still working on this. Reverted to r to get things working for other
refactorings. About to get rid of it

briankung (approved):

briankung:
So you were right way back when, they were different enough to split into separate partials. Sorry about that - you were correct.

Theoretically right now, we have a way to attach comments to a comments div with `comments[data-commentable-id=123]`, and use css selectors to pick elements with or without a `reference` class depending on whether it is or isn't a reference. If we could guarantee that ids were unique, then we wouldn't need to clarify whether or not it had a reference class. But that's kind of how I would want to select a comments container to add comments to. The `comments` class says it's a container for comments, and the combination of `data-commentable-id` and `reference` allow for picking a unique comments container, even though a comment and a reference might have the same ID.

app/controllers/comments_controller.rb
```diff
@@ -11,11 +11,20 @@ def edit
...
     @comment = Comment.new(comment_params)
     @comment.user = current_user
-
     respond_to do |format|
       if @comment.save
         format.html { redirect_to :back, notice: 'Comment was successfully created.' }
         format.json { render :show, status: :created, location: @comment }
+        format.js do
+          reference = @comment.root.commentable
+          is_top_level = !!@comment.commentable
```
wsf246, line 20: !!!

app/controllers/comments_controller.rb
```diff
@@ -40,10 +50,12 @@ def update
   # DELETE /comments/1
   # DELETE /comments/1.json
   def destroy
+    reference = @comment.root.commentable
```
wsf246, line 53: might be worth doing a before_action for `reference = @comment.root.commentable`

app/helpers/comments_helper.rb
```diff
@@ -2,7 +2,12 @@ module CommentsHelper
   def comments_tree_for(comments)
     comments.map do |comment, nested_comments|
       render 'comments/comment', comment: comment do
-        content_tag(:div, comments_tree_for(nested_comments), class: 'reply') if nested_comments.size > 0
+        content_tag(
+          :div,
+          comments_tree_for(nested_comments),
+          class: "comments",
+          'data-commentable-id' => comment.id
```
wsf246, line 9: just out of understanding best practice, do you choose to create a custom data-id field instead of using the id field just to be specific?

briankung:
The ID field in HTML should be unique. Since I can't guarantee it's unique,
I am using a more explanatory field/field name

briankung:
Sure

app/views/comments/create.js.erb
```diff
@@ -0,0 +1,14 @@
+<%
+  comment_and_comments_container = render('comments/comment', comment: @comment) do
+    render('form', comment: @comment.children.build)
+    content_tag(:div, '', class: 'comments', 'data-commentable-id' => @comment.id)
```
wsf246, line 4: fancy

briankung:
"Sure" was supposed to be a response for the before action suggestion

wsf246:
Gotcha, going back to what you were saying before.  So best practice, go
with just the ID field, but if you can't guarantee uniqueness, create a
data ID field explicitly.

app/controllers/comments_controller.rb
wsf246: nevermind, it's not worth doing.

app/views/comments/edit.js.erb
```diff
@@ -0,0 +1,16 @@
...
+
+commentContent.html(newForm);
+
+commentSubmitButton = commentContent.find(".btn")
+commentSubmitButton.after(cancelLinkHtml)
+
+cancelLink = commentSubmitButton.next().find(".cancel-link");
+cancelLink.on('click', function(e) {
+  e.preventDefault();
+  commentContent.html(undoContent);
```
wsf246, line 15: sweet

app/controllers/comments_controller.rb
briankung: This is a trick to cast something into a boolean

PR #129, opened 2016-10-21
Crossref search

briankung:
**Problem:** Adding papers sucks

**Solution:** Let users search crossref

**Complications:** 
- Crossref search also sucks. Waiting on answers to some questions:
  - [x] Can we limit search results to journals only?
  - [ ] For a single work result (https://api.crossref.org/works/10.1111/vec.12510), can we assume that the first author in the authors array is the first author?
  - [x] ~~What are coins? (2nd to last key: https://search.crossref.org/dois?sort=score&q=renear+palmer)~~
  - [ ] What should we do if we notice that your documentation is incorrect / out of date
- I'm removing duplicate titles, but they may not actually be duplicates

**Changelog:** 
- Add a search bar
- Visual indicators of status
- Add reference by search result DOI
- Some defensive coding against bad submissions

### DIFF
--- a/app/assets/javascripts/references.js
+++ b/app/assets/javascripts/references.js
@@ -16,7 +16,7 @@ $(document).ready(function() {
     if (this.value === '') {
       hideResultsAndClearSubmittable();
     } else {
-      $.get('https://search.crossref.org/dois?sort=score&q=' + this.value).done(function(data) {
+      $.get('https://search.crossref.org/dois?sort=score&type=Journal+Article&rows=10&q=' + this.value).done(function(data) {
         var results;
 
         if (data.length > 0) {
--- a/app/assets/javascripts/references.js
+++ b/app/assets/javascripts/references.js
@@ -2,9 +2,13 @@ $(document).ready(function() {
   var showCrossrefResults = debounce(function() {
     var resultsElement = $('#crossref-results');
 
-    if (this.value === '') {
+    var hideResults = function() {
       resultsElement.toggleClass('hidden', true);
       resultsElement.children().remove();
+    }
+
+    if (this.value === '') {
+      hideResults();
     } else {
       $.get('https://search.crossref.org/dois?sort=score&q=' + this.value).done(function(data) {
         var results, resultsByFullCitation;
@@ -20,10 +24,12 @@ $(document).ready(function() {
           });
 
           resultsElement.children().remove();
-          results = $.map(resultsByFullCitation, function(citation) {return "<li class='list-group-item'>" + citation + "</li>"})
+          results = $.map(resultsByFullCitation, function(citation) {return "<li class='list-group-item'>" + citation + "</li>"});
           resultsElement.append(results.join(''));
           resultsElement.toggleClass('hidden', false);
-        }
+        } else {
+          hideResults();
+        };
       })
     }
   }, 200);
--- a/app/assets/javascripts/references.js
+++ b/app/assets/javascripts/references.js
@@ -10,24 +10,24 @@ $(document).ready(function() {
 
     if (this.value === '') {
       // When no selection made, hide the locator and the paper title fields
-      if (!paperLocator.hasClass('hidden')) { paperLocator.addClass('hidden'); };
+      paperLocator.toggleClass('hidden', true)
       if (!paperLocator.prop('disabled')) { paperLocator.prop('disabled', true); };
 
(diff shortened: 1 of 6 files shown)


briankung:
Here's a URL for testing: https://afternoon-meadow-22528-pr-129.herokuapp.com/testuser/maintaining-mobility-in-old-age

Note that you can't actually add a paper through the search results at the moment. Also, don't hit the "add paper" button. I don't know what it does.

briankung:
It just occurred to me that Crossref may not just be searching papers.

app/assets/javascripts/references.js
```diff
@@ -1,4 +1,61 @@
...
+    hideResults();
+    $('.crossref').toggleClass('submittable', false);
+    $('#crossref-locator-id').val('');
+  }
+
+  var showCrossrefResults = debounce(function() {
+    if (this.value === '') {
+      hideResultsAndClearSubmittable();
+    } else {
+      $.get('https://search.crossref.org/dois?sort=score&q=' + this.value).done(function(data) {
```
wsf246: you can also search using the API
http://api.crossref.org/funders/100000015/works?query=global+state&filter=has-orcid:true

was it easier to use search.crossref?

wsf246:
The API doc is on the github:
https://github.com/CrossRef/rest-api-doc/blob/master/rest_api.md

You should be able to limit search results and limit to journal articles by using the filter.  You can do it on search.crossref too with type "type=Journal+Article".  You can use filter type as well for the API, though I'm still looking into the exact method

wsf246:
For the API, rows limits the results and filter=type:journal-article limits it to journal articles.

http://api.crossref.org/works?query=meditation&filter=type:journal-article&rows=5

wsf246:
I think pubmed grabbed the wrong paper if you try to get a crossref paper with a doi through a query instead of using the DOI as the specific search criteria

Here is the a DOI searched in a query:
http://api.crossref.org/works?query=10.1016/j.neulet.2016.10.009
"C2-Alkylation of Pyridine N-Oxides with 1,1-Diborylalkanes"

And here it is searched directly with the DOI
http://api.crossref.org/works/10.1016/j.neulet.2016.10.009
Religious and spiritual importance moderate relation between default mode network connectivity and familial risk for depression

I think this explains why that abstract was wrong for that 1 paper.

wsf246:
While going through the crossref stuff, there are some additional metrics that we may want to eventually grab:
-funders
-grants
-reference counts
-author affiliations
-copyright?

We may also want to expand search beyond just journal articles.  I would consider conference papers and reports as well, though I'm not 100% sure what reports are, could be more opinion based.

wsf246:
I double checked the authors across some pubmed examples and they appear to be in the same order.  Pubmed I know orders them correctly.

I think for all intents and purposes, the Crossref REST API is the most up to date resource.  I can confirm both these things with the crossref people.

app/assets/javascripts/references.js
briankung: I think search.crossref.org was the search endpoint you gave me, or the one in `crossref.rb`

briankung:
> I think this explains why that abstract was wrong for that 1 paper.

That doesn't sound quite right. The original paper was something about Bruce Lee. Click the abstract here: https://afternoon-meadow-22528-pr-129.herokuapp.com/testuser/maintaining-mobility-in-old-age/19

> I double checked the authors across some pubmed examples and they appear to be in the same order. Pubmed I know orders them correctly.

We shouldn't program by coincidence, though. I'd rather they confirm independently in writing or in the API docs that this is the case.

> The API doc is on the github...I think for all intents and purposes, the Crossref REST API is the most up to date resource. I can confirm both these things with the crossref people.

I meant their API documentation is out of date or wrong, not that the API itself was. I found broken endpoints or things that just didn't function.

> While going through the crossref stuff, there are some additional metrics that we may want to eventually grab

GitHub hasn't been showing my emailed comments, but this is a good reason to use a JSONB field to capture the json results. We may not want to use these metrics right away, but if we capture the query results, we can add them whenever we want.

> For the API, rows limits the results and filter=type:journal-article limits it to journal articles.

Cool, thanks!

briankung:
Another thing that would be nice is knowing how to search for `filter=type:journal-article` OR `filter=type:dissertation`. The changelog says they support `OR` operations. The documentation just doesn't say how to do it. That's kind of what I mean by being out of date.

briankung:
Exactly why I want to use JSONB. But probably in a different pull request

briankung:
Cool, thanks!

PR #130, opened 2016-10-22
Allow linebreaks in comments

briankung:
**Problem:** Can't see linebreaks in comments

**Solution:** Show line breaks

**Complications:** Not really sure what else `#simple_format` does...

**Changelog:** 
- Allow linebreaks with `#simple_format`


app/views/comments/_comment.html.erb
```diff
@@ -3,7 +3,7 @@
     Added by <strong><%= link_to comment.user.username, user_profile_path(comment.user.username) %></strong> on <%= l(comment.created_at, format: '%B, %d %Y %H:%M:%S') %>
   </p>
 
-  <p><%= comment.content %></p>
+  <p><%= simple_format(comment.content, {}, sanitize: true) %></p>
```
wsf246, line 6: _thumbs up_

PR #132, opened 2016-10-25
Users references relation

briankung:
**Problem:** References weren't capturing data about who created them

**Solution:** 
- [x] Capture and backfill reference owner data
- [x] Mention user's username on their lists index page
- [x] Sort references by votes, then creation date instead of votes then publish date


app/assets/stylesheets/_crossref_search.scss
```diff
@@ -18,6 +18,7 @@
   span.input-group-addon {
     background-color: $brand-primary;
     color: white;
+    cursor: pointer;
```
wsf246, line 21: fancy

app/views/lists/index.html.erb
```diff
@@ -1,5 +1,5 @@
 <h1>
-  <span>Lists</span>
+  <span><%= "#{@user.username}'s" if @user %> Lists</span>
```
wsf246, line 2: _thumbs up_

db/migrate/20161025201812_add_user_to_references.rb
```diff
@@ -0,0 +1,5 @@
+class AddUserToReferences < ActiveRecord::Migration[5.0]
+  def change
+    add_reference :references, :user, foreign_key: true
```
wsf246, line 3: didn't know this was a thing.

wsf246:
We should consider adding "added by xxxx" to each reference on a list, or maybe only if it wasn't added by the creator of the list.  Otherwise LGTM.

db/migrate/20161025201812_add_user_to_references.rb
briankung: This is generated by `rails generate migration AddUserToReferences user:references`

And yeah, you could just do an integer, too. I'm not sure what the `foreign_key: true` bit does

app/views/lists/index.html.erb
briankung: If you hit colon : you can start typing and get a emoji selector 😆  thumbs up is : +1:

app/assets/stylesheets/_crossref_search.scss
briankung: Glad you liked it 👍

PR #133, opened 2016-10-28
Removes apostrophes from slugs

briankung:
**Problem:** Slugs should not replace apostrophes with dashes

### DIFF
--- a/app/models/list.rb
+++ b/app/models/list.rb
@@ -21,7 +21,11 @@ class List < ApplicationRecord
             }
 
   def to_slug
-    self.name.downcase.gsub(/[^\p{N}\p{L}]/, '-').gsub(/-{2,}/, '-')
+    self.name
+      .downcase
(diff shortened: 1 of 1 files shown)


app/models/list.rb
```diff
@@ -21,7 +21,11 @@ class List < ApplicationRecord
             }
 
   def to_slug
-    self.name.downcase.gsub(/[^\p{N}\p{L}]/, '-').gsub(/-{2,}/, '-')
+    self.name
+      .downcase
+      .gsub("'", '')                # Remove apostrophes
+      .gsub(/[^\p{N}\p{L}]/, '-')   # Replace non-number, non-letter characters with a dash
+      .gsub(/-{2,}/, '-')           # Combine any contiguous dashes
```
wsf246, line 28: we should break this up into multiple PRs 
j/k, merged!

PR #134, opened 2016-10-28
WIP - Search improvements

briankung:
**Problem:** Noticed some UX hiccups when Andrew was testing the site. The search bar was hard to notice and then it wasn't immediately obvious what to click.

**Solution:** Do away with the unnecessary step of clicking "add paper" and put the button on the left

**Complications:** 
1. I think I can combine the search and add paper partials. The DOI/PMID/etc options will be on the right as a dropdown.
2. I think this is one of those weird moments when removing a step actually makes it feel slower. Clicking the button to add the paper was strangely satisfying. `¯\_(ツ)_/¯`

**Changelog:** 
- Refactor JS
- Shuffle add paper around to the left side
- Submit paper on clicking the result


app/assets/javascripts/crossref.js
```diff
@@ -0,0 +1,60 @@
...
+
+  $('#crossref-search').on('input', showCrossrefResults);
+  $('#crossref-search').on('input', hideResultsAndClearSearchValue);
+  $('#crossref-search').on('keydown', function (e) {
+    if (e.keyCode === 27) {
+      this.value = '';
+      hideResultsAndClearSearchValue();
+    };
+  });
+});
```
wsf246, line 60: fun

app/assets/javascripts/references.js
```diff
@@ -1,61 +1,4 @@
...
-
-  $('#crossref-submit').on('click', function(e) {
-    if ($(this).parent().hasClass('submittable')) {
-      $(this).siblings('form#new_reference').submit();
-    };
-  });
-
   $('a.delete-reference').on('click', function(e) {
     e.preventDefault();
     $(e.target).siblings('form').submit();
```
wsf246, line 4: _thumbs up_

app/views/crossref/_search.html.erb
```diff
@@ -1,6 +1,6 @@
 <div class="input-group crossref">
-  <%= text_field_tag 'crossref-search', nil, class: 'form-control', placeholder: 'Search for a paper' %>
   <span class="input-group-addon" id="crossref-submit">add paper</span>
+  <%= text_field_tag 'crossref-search', nil, class: 'form-control', placeholder: 'Search for a paper' %>
```
wsf246, line 3: the UI changes that will be the recurring stories we tell in the book.

PR #136, opened 2016-10-29
WIP - List Show Formatting

wsf246:
**Problem:** 
Fix edit route and other UI to make site generally more usuable

**Solution:** 
-fix edit route
-update placement of buttons
-update references tags

**Complications:** 

**Changelog:**


config/routes.rb
```diff
@@ -23,5 +23,6 @@
   get ':username' => 'users/lists#index', as: :user_profile
   get ':username/:list_slug' => 'users/lists#show', as: :user_list
   get ':username/:list_slug/edit' => 'lists#edit', as: :edit_user_list
+  get ':username/:list_slug/destroy' => 'lists#destroy', as: :destroy_user_list
   get ':username/:list_slug/:reference_id' => 'users/references#show', as: :user_reference
```
briankung: Once we figure out what these routes are, we should definitely use the `resources` shortcuts, but this is fine for now

briankung:
Looks good so far

briankung:
What's left to do?

wsf246:
I'm playing w some formatting, lemme push what I have so far

wsf246:
I'm working on the search and add locator right now, trying a view
different ways, so the way it looks now is temporary.  Just trying some UI
changes to see if I can make it more intuitive.

app/controllers/lists_controller.rb
```diff
@@ -62,7 +62,7 @@ def destroy
   private
     # Use callbacks to share common setup or constraints between actions.
     def set_list
-      @list = List.find_by(slug: params[:id])
+      @list = List.find_by(slug: params[:list_slug] || params[:id])
```
briankung: We should really pick one type of ID and stick to it whenever possible.

app/views/references/_reference.html.erb
```diff
@@ -9,7 +9,9 @@
     <% else %>
       <h4 class="paper-title">
         <%= link_to paper.title, user_reference_path(reference.list.user, reference.list, reference) %>
-        <small>contributed by <%= link_to reference.user.username, user_profile_path(reference.user) %></small>
+        <% if reference.user_id != reference.list.user_id %>
```
briankung: I don't think this should be conditional.

briankung:
I'm not a huge fan of this:
screen shot 2016-11-01 at 11 44 44 am

I was also going to redo this myself on the earlier WIP branch:
screen shot 2016-11-01 at 11 44 48 am

---

I think, overall, we might want to hold off on this until we can talk about it.

briankung:
And/or get design feedback on it.

From a technical standpoint, I agree with choosing a locator type, I was going to approach it way differently.

app/controllers/lists_controller.rb
wsf246: yea I agree.   The routing syntax is confusing to me and I wasn't sure how to get destroy to pass :list_slug instead of params[:id].  Either way though, do you want to stick with :list_slug as the identifier?

app/views/references/_reference.html.erb
wsf246: I'm playing around with some formatting so most of the rest of the stuff besides the route is just to give a general facelift.  For this, i'm generally concerned about overloading the page with text, especially once we have tags for each of the papers and especially if we do end up deciding to go with methodologies and biases.  Plus an other indicators we might want to show up front like reference count.  

My thinking here is to reduce repetitive text and for this, we can have the reader assume any non-marked reference is contributed by the creator of the list.  

Again, these are just putting some thoughts down to visualize and play around with so let's talk through these.

wsf246:
yea, let's talk through these changes.  my focus here is to simplify the UI and go with what felt intuitive to me.  I think it's also beneficial to do some usability tests with this and we could do a pseudo A/B test with what you have in mind.

wsf246:
This should be good to go for now, let me know what you think.

config/routes.rb
```diff
@@ -21,7 +21,8 @@
   end
 
   get ':username' => 'users/lists#index', as: :user_profile
-  get ':username/:list_slug' => 'users/lists#show', as: :user_list
-  get ':username/:list_slug/edit' => 'lists#edit', as: :edit_user_list
-  get ':username/:list_slug/:reference_id' => 'users/references#show', as: :user_reference
+  get ':username/:id' => 'users/lists#show', as: :user_list
+  get ':username/:id/edit' => 'lists#edit', as: :edit_user_list
+  get ':username/:id/destroy' => 'lists#destroy', as: :destroy_user_list
+  get ':username/:id/:reference_id' => 'users/references#show', as: :user_reference
```
briankung: this one should be `:list_id` and `:reference_id` should be `:id` according to rails conventions

app/views/lists/show.html.erb
```diff
@@ -1,21 +1,26 @@
...
-
-<p>
-  <strong>Tags:</strong>
-  <%= @list.tag_list %>
+  <% @list.tag_list.each do |tag| %>
+    <a class="btn btn-default btn-xs"><%= tag %></a>
+  <% end %>
 </p>
+<strong>Description</strong>
+<%= simple_format(@list.description) %>
```
briankung: This should have the same sanitize options that the other uses of simple_format do

app/views/lists/_add_reference.html.erb
```diff
@@ -1,25 +1,30 @@
...
+                          disabled: true %>
 
-      <%= p.text_field  :title,
-                        placeholder: 'Paper title',
-                        class: "form-control hidden paper-title",
-                        disabled: true %>
+        <%= p.text_field  :title,
+                          placeholder: 'Required - Paper Title ex: Regulation of the Neural Circuitry of Emotion',
+                          class: "form-control hidden paper-title input-sm",
+                          size: 70,
```
briankung: Try CSS. I think `text-overflow: wrap;` or something like that might work. https://davidwalsh.name/placeholder-overflow

briankung:
Merge conflicts :/

PR #137, opened 2016-10-30
Refactor / Improve Pubmed Import process

briankung:
**Problem:** With `pubmed.rb` as it existed earlier, it was very action-specific instead of representing a toolkit, and it was difficult to persist the imported data.

**Solution:** Refactor and add some features

**Complications:** This particular feature is ready to go, but I'll do these in separate pull requests:
- Look over the DOI import process
- Improve the import by url process
- Improve checks for locator parameters - probably move them into the locators

Oh yeah, I forgot this TODO:
- [x] Ensure the import doesn't break on pmid 123 (which has an ampersand that I unescape, which breaks the XML parser) or other weird cases I've noted
- [x] Make everything default to XML
- [x] Return a request object that we can grab the uri from to save it to ImportApiResponse

**Changelog:** 
- Changes locator API
- Changes Pubmed API
- Persists imported data
- Breaks Pubmed into several classes
- Guard against DOI not found if searching Pubmed by DOI
- Persists response data to `ApiImportResponse`
- Switches to Nokogiri for all the things


db/schema.rb
```diff
@@ -90,12 +90,13 @@
...
     t.date     "published_at"
-    t.datetime "created_at",   null: false
-    t.datetime "updated_at",   null: false
+    t.datetime "created_at",       null: false
+    t.datetime "updated_at",       null: false
     t.text     "abstract"
     t.string   "doi"
     t.string   "pubmed_id"
     t.string   "publication"
+    t.jsonb    "data_from_import"
```
wsf246: _thumbs up_

briankung:
Ugh. This is good to go. Still have to update DOI import and Link import

briankung:
Shit, I forgot to update one thing to not use `Nokogiri`. `Hash.from_xml` breaks when there are ampersands

wsf246:
looks like it still accepts faulty pubmed IDs.  I tried to add a DOI using pubmed ID "10.1007/springerreference_44360" and it grabbed the wrong pubmed ID.  Just a headsup.

briankung:
DOI import is unaffected. That's another issue

wsf246:
Right I mean I can still get faulty Pubmed IDs in.  For example if I use
Pubmed import for 10.1007/springerreference_44360, I get a paper from
pubmed.
https://www.ncbi.nlm.nih.gov/pubmed/?term=10.1007%2Fspringerreference_44

Additionally, I completely wrong pubmed IDs break the site when it tries to
parse the XML on the pubmed API error screen
https://eutils.ncbi.nlm.nih.gov/entrez/eutils/efetch.fcgi?db=pubmed&rettype=abstract&id=poo

briankung:
How are you doing this? Provide steps to reproduce.

If you're selecting DOI in the add paper by drop down, that's still the old
system.

briankung:
You shouldn't be selecting Pubmed import for a DOI, for instance

briankung:
Anything DOI related is probably as broken as before, if not more

wsf246:
I'm selecting Pubmed ID and putting bad data:
image

It breaks on import:
image

briankung:
I'm not surprised, I preferred breaking. What did you mean by the DOI bit?

briankung:
I'll fix it before moving onto the DOI stuff

wsf246:
I'm trying to recreate but I'm having issues destroying the paper I created:
image

`ActiveRecord::InvalidForeignKey: PG::ForeignKeyViolation: ERROR:  update or delete on table "papers" violates foreign key constraint "fk_rails_80c22e88b0" on table "api_import_responses"
DETAIL:  Key (id)=(132) is still referenced from table "api_import_responses".`

briankung:
Huh, the foreign key thing is a surprise. I didn't know it maintains
consistency by preventing deletion if it's a foreign key.

wsf246:
yea, do you think we should do a dependent destroy on the import if the paper gets destroyed or vice versa?  there's not way for a user to destroy a paper for now so it would just be for us on the back end.

briankung:
I feel like we should keep imports. I could be persuaded either way

wsf246:
So for the pubmed API, if I accidentally try to import a DOI using the pubmed ID 
image

There's a chance it adds the wrong paper, though for some reason I can't get it to do it now.
This is what the API gets:
https://eutils.ncbi.nlm.nih.gov/entrez/eutils/efetch.fcgi?db=pubmed&rettype=abstract&id=10.1007/springerreference_61489

but I'm getting this error now
image

PR #143, opened 2016-11-09
Fix DOI Import

briankung:
**Problem:** I broke DOI import

**Solution:** Fix DOI import

**Complications:** None, hopefully now?

Well, okay, some minor things changed, too. I'm now using the print publish date instead of the electronic crossref record's create date for the published_at date. I'm also using the journal title for the publication field instead of the publisher. Unfortunately those are confusingly named.

I'm also skipping validations on associated authors and links so I can find or create them based on their attributes.


app/models/paper.rb
```diff
@@ -1,16 +1,16 @@
...
+
   acts_as_taggable
   acts_as_taggable_on :biases, :methodologies
 
   has_and_belongs_to_many :authors
   has_many :lists, through: :references
   has_many :references, dependent: :destroy
   has_many :links, dependent: :destroy
-  has_many :api_import_responses
+  has_many :api_import_responses, dependent: :destroy
```
wsf246, line 11: looks like you chose for dependent destroy, probably the wise thing to do for now until we have a good reason to save them.

app/models/paper.rb
```diff
@@ -55,4 +55,20 @@ def valid_methodologies
       'case report'
     ]
   end
+
+  # Skips validations, fallthrough to autosave methods for easy import
+  def validate_associated_records_for_authors; true; end
+  def validate_associated_records_for_links; true; end
```
wsf246, line 61: for the other branch I'm working on, I'm letting people add links, this might conflict with this validation, but I'll see if there is any issues post merge.

wsf246: conflict with this non-validation that is.

app/models/paper.rb
briankung: Right, out of development convenience, mostly. We have to vet these dependent destroy relationships closely in prod and see what people have access to.

app/models/paper.rb
briankung: Add links when they create a paper, correct? If it's nested in the paper, that would cause a problem. I was experimenting with a way to tell the validations if it was an import or not, but it got too involved for this PR.

wsf246: cool yea.

app/models/services/crossref/resource.rb
```diff
@@ -0,0 +1,52 @@
+class Crossref
+  class Resource
+    attr_accessor :id, :response
+
+    def initialize id
+      self.id = id.to_s
+      endpoint = URI.parse("http://api.crossref.org/works/#{self.id}")
```
wsf246: do you need to convert it to a string and then do interpolation?

wsf246: also wondering if it would be worthwhile to split out this html out with the html you're using for the crossref autofill similar to what you did for pubmed.  Now that I'm thinking outloud, maybe not because you cant grab the html for javascript unless you use ujs.

app/models/services/crossref/resource.rb
```diff
@@ -0,0 +1,52 @@
...
+        memo
+      end
+    end
+
+    def mapper
+      {
+        title:              lambda {|data| data.dig 'message', 'title', 0 },
+        publication:        lambda {|data| data.dig 'message', 'short-container-title', 0 },
+        doi:                lambda {|data| self.id },
+        pubmed_id:          lambda {|data| Pubmed.get_uid_from_doi(id) },
```
wsf246, line 34: sweet.  here's the documentation for the id converter if you're interested:
https://www.ncbi.nlm.nih.gov/pmc/tools/id-converter-api/

app/models/services/pubmed/resource.rb
```diff
@@ -33,40 +22,24 @@ def map_attributes mapper, data
...
-        pubmed_id:          lambda {|data| data.xpath('//PMID').text },
-        abstract:           lambda {|data| data.xpath('//AbstractText').map(&:text).join("\n\n") },
-        published_at:       lambda {|data| Date.parse data.xpath('//PubDate').map(&:text).join(' ') },
-        authors_attributes: lambda {|data| authors = data.xpath('//AuthorList/Author')
+        title:              lambda {|data| data.css('ArticleTitle').text },
+        publication:        lambda {|data| data.css('Journal Title').text },
+        doi:                lambda {|data| data.css('ArticleId[IdType=doi]').text },
+        pubmed_id:          lambda {|data| data.css('PMID').text },
+        abstract:           lambda {|data| data.css('AbstractText').map(&:text).join("\n\n") },
+        published_at:       lambda {|data| Date.parse data.css('PubDate').map(&:text).join(' ') },
```
wsf246, line 30: sweet.  good to know how css vs xpath works.

app/models/services/crossref/resource.rb
briankung: No

briankung: This doesn't respond with HTML, though

app/models/services/crossref/resource.rb
wsf246: same comment I had yesterday here about checking if a pubmed ID already exists in our database but sounds we decided that will take place in the API imports meshing somewhere.

app/models/services/pubmed/resource.rb
briankung: `.css` is a convenience method that converts the css-style selector to an xpath selector. I couldn't figure out how to do `ArticleId[IdType=doi]` in xpath super easily so I converted it to `.css` instead. I figure we're already used to using CSS selectors, so why not just use it for XML?

app/models/services/pubmed/resource.rb
```diff
@@ -1,25 +1,14 @@
...
-      else
-        self.response = nil
-      end
+      self.response = get_response(id)
     end
 
     def paper_attributes
-      @paper_attributes ||= map_attributes mapper(), Nokogiri::XML(response.try(:body))
+      @paper_attributes ||= map_attributes mapper(), Nokogiri::XML(response.try(:body), &:noblanks)
     end
```
wsf246, line 12: cool no blanks.

app/models/services/crossref/resource.rb
wsf246: sry, I meant https

app/models/services/pubmed/resource.rb
briankung: I also think it might be able to optimize the xpath query a bit, because I don't know how to use xpath. Apparently using `//SoAndSo` is inefficient because it searches the whole document for the node. If I really wanted it to be fast, I would use an absolute path like I had before - `//ArticleSet/Article/etc...` - but it's just so ugly to look at...

app/models/services/crossref/resource.rb
wsf246: I guess you're using it for other things then?

wsf246: I was confused because it seemed redundant to do id.to_s and the use the string interpolation, but I assume you're using ID somewhere else.

app/models/services/crossref/resource.rb
briankung: Cool, I might add this in another PR

app/models/services/crossref/resource.rb
briankung: Oh yeah, that makes a lot of sense. Let's do that

briankung: Oh, I see what you mean now. Good point. I think I need it more in the `Pubmed::Resource` class. Good catch

briankung: Turns out we do use that later on, woops

wsf246:
This looks good to go.  Though as this meant to fix the add reference bit as well.  When I try to add a doi using the "add paper by: DOI", the site breaks:

image

Or it says parameters can't found.

app/models/services/crossref/resource.rb
```diff
@@ -33,8 +33,14 @@ def mapper
...
-          date = data.dig('message', 'published-print', 'date-parts', 0)
-          Date.parse(date.join('/')) if date
+          if (date_parts = data.dig('message', 'published-print', 'date-parts', 0))
+            year, month, day = date_parts
+
+            return nil unless year
+
+            date = "#{year}/#{month || 1}/#{day || 1}"
+            Date.parse(date)
+          end
```
wsf246, line 43: fun

PR #144, opened 2016-11-11
Pubmed refactor

briankung:
Just cleaning up the Pubmed service a bit

### DIFF
--- a/app/models/services/pubmed/http.rb
+++ b/app/models/services/pubmed/http.rb
@@ -1,6 +1,6 @@
 class Pubmed
   class Http
-    def self.default_parameters
+    def self.eutils_params
       {
         db: 'pubmed',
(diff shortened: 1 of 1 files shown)


app/models/services/pubmed/http.rb
```diff
@@ -1,33 +1,39 @@
...
-    def self.esearch options = {}
-      uri = generate_uri("/entrez/eutils/esearch.fcgi", default_parameters.merge(options))
-
-      Net::HTTP.get_response(uri)
+    def self.id_params
+      {
+        tool: 'projectcredo',
+        email: 'accounts@projectcredo.com'
+      }
     end
```
briankung, line 16: Note that these identifiers are things that the various Pubmed services expect to have for their own instrumentation purposes.

PR #147, opened 2016-11-11
change autosave assocaited records for links from name to url

wsf246:
Quick fix, didn't notice this before

- Enter high level description of changes here.


briankung:
This should have been fixed already

PR #148, opened 2016-11-12
Edit paper

wsf246:
**Problem:** 
User has no ability to edit un-imported data for the paper including abstract, links, and tags.
**Solution:** 
Provide forms to edit the paper on the list show page
**Complications:** 
Some of this I think can be improved with Ajax and therefore move the javasript I wrote into ujs stuff, but for now I think this is fine. 

I also still need to add an abstract-editable attribute and link-editable attribute but I decided to cut this PR here for review and do a separate PR for that.  The next PR will allow the user to edit the abstract if it wasn't imported and remove links. 

**Changelog:** 

- Added Abstract form for papers with no abstracts
- added links form
- added tags form 
- added tags javascript to toggle the form.


briankung:
Looks good except that there's some unexpected behavior with regards to adding links:

If you add a link to a paper, you can "steal" links from other papers. Add "https://google.com" to one paper and then another through the links form. You can see that it transfers from one paper to another - in my case, it actually took a link from a paper I added by URL, which means that there's no canonical identifying information for that paper.

wsf246:
wooooahhh, trippy

app/views/papers/_tags_form.html.erb
```diff
@@ -0,0 +1,10 @@
+<div class="tags-form hidden">
+  <%= form_for(paper, html: {class: 'form-inline'}) do |f| %>
+    <div class="form-group col-md-12">
+      <%= f.text_field :tag_list, value: paper.tag_list.join(","), placeholder: "No tags added yet...", class: "form-control input-sm" %>
+      <%= f.submit 'Add tags', class: "btn btn-xs btn-primary" %>
+      <%= link_to "cancel", '#', class: "cancel-add-tag link-btn" %>
+    </div>
+  <% end %>
+  <i class='help-block'><small>Tags should be seperated by commas.</small></i>
+</div>
```
briankung, line 10: It would be cool if we could do the button thing for tags as you write them, too. I think StackOverflow does this. Just a bonus, though.

app/views/papers/_tablist.html.erb
```diff
@@ -28,13 +29,17 @@
...
     <% end %>
+
+    <%= form_for(paper, html: {class: "form-inline"}) do |p| %>
+    <div class="form-group">
+      <%= p.fields_for :links, paper.links.build do |l| %>
+        <%= l.text_field :url, class: "form-control", size: 50%>
+      <% end %>
+      <%= p.submit 'Add Link', class: "btn btn-xs btn-primary" %>
+    <% end %>
+    </div>
```
briankung: You don't have to do this, but tablist has pissed me off for a while. I was thinking we should just take its component pieces and embed them in the reference partial instead of adding another layer of indirection. We don't reuse the tablist anywhere, either, so...

app/models/paper.rb
```diff
@@ -66,9 +67,4 @@ def autosave_associated_records_for_authors
     end
   end
 
-  def autosave_associated_records_for_links
-    self.links = links.map do |link|
-      Link.find_or_create_by name: link.name
-    end
-  end
```
wsf246, line 73: here's my solution.  I think we can get rid of this since we are checking for existing links for the link locator type.  I added a validation to the model so it currently only allows a link once for a paper.  We can make the validation unique by URL, but I figure this will probably be an edge case that we can always fix in postgres later.

I'm a little more concerned that a uniqueness validation for URL could break on an import if the URLs are not unique in crossref.

briankung: I think this is fine.

app/views/papers/_tags_form.html.erb
wsf246: you mean like this?

image

This is definitely where I'm hoping this will end up.  I think that would involve some more ajaxing so I will pass on this for now.  I really like the definitions and autocomplete feature though.  I'm also still keen on the idea of colors for methodology and biases too.  I think the colors would stay consistent across the site (blue for methodology, red for biases) but people could choose which it is for the tag.  Just a thought.

briankung: Yeah, like that, but dumber. We'd basically just split it and create a new button whenever someone hits the space bar or the comma button

briankung: We'd eventually want to pull in tags from the backend, but not for now.

app/models/paper.rb
```diff
@@ -12,6 +12,7 @@ class Paper < ApplicationRecord
 
   accepts_nested_attributes_for :authors, reject_if: proc { |attributes| attributes['name'].blank? }
   accepts_nested_attributes_for :links
+  validates_associated :links
```
briankung, line 15: In my testing, I think this wasn't actually necessary, but your mileage may vary

wsf246: so I'm doing all of the updating of these paper attributes through papers#update.  I have the validates associated for the new links so it validates its a URL and checks for uniqueness.

briankung: I understand, I just meant that in my testing, the validation was done whether or not this line was here.

wsf246: ah ok.  If this is good to merge I can merge.

briankung: Yeah, I think so

PR #152, opened 2016-11-16
Editable fields

wsf246:
**Problem:**
User should be able to add abstracts and links but only for papers that are not imported
**Solution:**
create boolean fields for paper, abstract, and links to show if the fields can be editted
**Complications:**
I created 3 boolean fields to show editable that get flipped during the import process.  They all default to true, this seemed like the logical way to set it up.  I added paper_editable, which we could use for any papers that are not imported.  Those papers will have an edit button that will direct them to the paper form.  I'm going to do that in a later PR.

I should also add a migration to flip the editables to false.

I added some ajax, but minimally.  Some additional jquery would make the experience less choppy, but this should be ok for now.  

There are a good number of things we should talk through as well including:
- where the editable toggle should sit.  I currently have it in the #find_or_import_paper methods in the respective locator right before the paper creates.
-how much ajax we should use
-some styling questions and placement of links

**Feature Changelog:**

- Added boolean fields
- made abstract editable
- made links removable
- added ajax
- moved permalink to its own link
- set title to go to direct link of the paper


app/models/services/locators/doi_paper_locator.rb
```diff
@@ -15,6 +15,7 @@ def find_or_import_paper
 
     if paper_attributes
       response = crossref.resource.response
+      (paper_attributes[:links_attributes].map {|l| l[:link_editable] = false}) if paper_attributes[:links_attributes].present?
```
briankung: We should probably put this in the mapper. Ditto with the other one.

briankung: I was going to do something similar to turn the validations on and off with a virtual attribute (like `locator_type` and `locator_id` on `Paper`) called `imported`. If `imported` was true, I was going to turn off the validations or something like that. I can't remember exactly what my plan was, now.

config/routes.rb
```diff
@@ -20,9 +20,13 @@
     resource :vote, controller: 'comments/votes', only: [:create, :destroy]
   end
 
-  get ':username' => 'users/lists#index', as: :user_profile
+  get ':reference_id/:id/remove' => 'links#destroy', as: :destroy_link
+  get ':username/:list_id/:id/edit_abstract' => 'references#edit_abstract', as: :edit_abstract
```
briankung: These two paths should be made RESTful. The path for a destroy action should be a `DELETE` on the id, and editing a reference should happen on the reference path through an `UPDATE` action

briankung (changes requested):

app/assets/javascripts/references.js
```diff
@@ -1,9 +1,18 @@
 $(document).ready(function() {
+  //Delete referene from a list
```
briankung: Let's remove comments that explain _what_ is happening and not _why_ it's happening. What is happening _should_ be evident from the code itself.

briankung (changes requested):

app/controllers/references_controller.rb
```diff
@@ -35,7 +35,20 @@ def destroy
     end
   end
 
+  # GET /username/references/1/edit_abstract
+  def edit_abstract
```
briankung: This should just be `edit`, even if we're only editing the abstract

app/views/papers/_links.html.erb
```diff
@@ -0,0 +1,28 @@
+<% paper = reference.paper %>
```
briankung: Since `links` have their own controller now, we should put link related views in their own view folder.

app/views/papers/_tablist.html.erb
```diff
@@ -14,35 +13,3 @@
...
-        <% end %>
-        <%= p.submit 'Add Link', class: "btn btn-xs btn-primary" %>
-        <i class="help-block"> Add a primary source link for this paper that represents the paper itself </i>
-      </div>
-    <% end %>
-  </ul>
-</div>
-<div class="paper-detail well collapse <%= "in" if show_comments %>" id="comments-<%= reference.id %>">
-  <%= render 'references/comments', reference: reference %>
-</div>
```
briankung, line 48: yayyy deleting code 👍

app/views/references/edit_abstract.js.erb
```diff
@@ -0,0 +1,7 @@
+newForm = "<%= escape_javascript(render('papers/abstract_form', paper: @reference.paper)) %>";
+abstractDiv = $("#edit-abstract-<%= @reference.id %>");
+abstractDiv.data("undo-form", abstractDiv.html());
```
briankung: Cool that this works.

briankung (changes requested):
Some minor stuff, overall looks pretty good, though

app/models/services/pubmed/resource.rb
```diff
@@ -27,6 +27,7 @@ def mapper
         doi:                lambda {|data| data.css('ArticleId[IdType=doi]').text },
         pubmed_id:          lambda {|data| data.css('PMID').text },
         abstract:           lambda {|data| data.css('AbstractText').map(&:text).join("\n\n") },
+        abstract_editable:  lambda {|data| false if data.css('AbstractText').present? },
```
briankung: I personally try to avoid negatives, so I'd rather `true if data.css('AbstractText').blank?`

wsf246: so I felt it would be easier for us to make all "_editable" fields default to true and flip it to false when importing.  Otherwise we'd have default be false and we'd have to flip it for every other type of incoming data.  That seems less defined.

briankung:
Hmm, I'd disagree. I think most of our papers would be imported, so we
should default to false.

This note, however, doesn't have to do with those defaults because they
will override the defaults regardless.

app/models/services/pubmed/resource.rb
wsf246: i think we'll also mostly import papers, but the issue is if we default to 'false', we'll need to make abstract_editable 'true' in all circumstances where we are not importing.  So right now, we'd need to add a line for crossref, link, and when pubmed doesn't have an abstract, vs just having 'false' for when we know it is importing an abstract.

I'm creating a default and making nulls 'true' to avoid 3 states the 'editables'.  That way we don't have to worry about a circumstance where we don't set 'editable'.  With the default, I can only make your switch suggestion if I also change the default.

config/routes.rb
```diff
@@ -20,9 +20,13 @@
     resource :vote, controller: 'comments/votes', only: [:create, :destroy]
   end
 
-  get ':username' => 'users/lists#index', as: :user_profile
+  get ':reference_id/:id/delete' => 'links#destroy', as: :destroy_link
```
briankung: It's the `get` that should change

config/routes.rb
```diff
@@ -20,9 +20,13 @@
...
+  get ':username/:list_id/:id/update' => 'references#edit_abstract', as: :edit_abstract
+
```
wsf246: double checking this is what you meant by making them RESTful.  Also, I wanted to make sure there isn't some path convention I'm messing up.  For the links#destroy, should I include username? I just included the params I thought we might want.

wsf246:
ah cool

config/routes.rb
```diff
@@ -20,9 +20,13 @@
     resource :vote, controller: 'comments/votes', only: [:create, :destroy]
   end
 
-  get ':username' => 'users/lists#index', as: :user_profile
+  get ':reference_id/:id/delete' => 'links#destroy', as: :destroy_link
+  get ':username/:list_id/:id/update' => 'references#edit_abstract', as: :edit_abstract
```
briankung: Instead of `get` it should say `patch` or `update` or both and the action should just be called `#edit`, unless you're actually updating the record, then it should be `update`

briankung: > You can use `get`, `post`, `put`, `patch` and `delete` methods to constrain a route to a particular verb. You can use the match method with the :via option to match multiple verbs at once:
> 
> `match 'photos', to: 'photos#show', via: [:get, :post]`

http://guides.rubyonrails.org/routing.html#http-verb-constraints

briankung: These are the routes that `resources :dogs` generates, which are RESTful in the Rails sense, and which we should try to emulate:

```
  Prefix Verb   URI Pattern              Controller#Action
    dogs GET    /dogs(.:format)          dogs#index
         POST   /dogs(.:format)          dogs#create
 new_dog GET    /dogs/new(.:format)      dogs#new
edit_dog GET    /dogs/:id/edit(.:format) dogs#edit
     dog GET    /dogs/:id(.:format)      dogs#show
         PATCH  /dogs/:id(.:format)      dogs#update
         PUT    /dogs/:id(.:format)      dogs#update
         DELETE /dogs/:id(.:format)      dogs#destroy
```

wsf246: so I swapped the link#destroy route with the resources rails method.  I was having trouble getting the delete verb to work properly, it kept mapping to the wrong route.  The resources method is probably better anyhow, especially if I add more actions.

For the edit_abstract, I kept it as a 'get' because it's just calling for the abstract form instead of updating.  I can switch the action to reference#edit, but I feel like that should be reserved for actually editing the whole reference instead of just the abstract.  

I read through http://guides.rubyonrails.org/routing.html but it didn't seem to give any advice about actions for editing a specific attribute.

This kind of gets at it but doesn't seem ideal: http://stackoverflow.com/questions/16952624/how-to-route-multiple-pages-to-edit-a-users-account-information-rails

app/controllers/references_controller.rb
wsf246: so should all edit functionality go through this action?  I guess I could also have the form pop up purely through javascript and not use ajax at all.  Let's talk through this since I clearly do not understand routing well enough. :-/

briankung: We'll talk about this, but look at the 7 standard RESTful routes: https://github.com/briankung/projectcredo/pull/152#discussion_r88381936

If you can accomplish what you need to through them, we should use one of the standard routes and actions. In this case, it would probably be `paper#edit`.

config/routes.rb
briankung: I forgot that it actually is an `edit` action, so it is a `get`. This should look like:

``` ruby
delete ':reference_id/:id' => 'links#destroy', as: :destroy_link
get ':username/:list_id/:id/edit' => 'paper#edit'
```

config/routes.rb
briankung: > For the edit_abstract, I kept it as a 'get' because it's just calling for the abstract form instead of updating

That's correct

> editing the whole reference instead of just the abstract

Quick question, we are editing the paper and not the abstract, correct?

> I read through http://guides.rubyonrails.org/routing.html but it didn't seem to give any advice about actions for editing a specific attribute.

That's because you can edit a single attribute through an action designed to edit the whole entity.

app/controllers/references_controller.rb
briankung: > not use ajax at all

Let's prefer this whenever possible.

config/routes.rb
briankung: > For the links#destroy, should I include username? I just included the params I thought we might want.

I think your initial assumption is correct

config/routes.rb
briankung: Oh. This is also not pointing at a link. Destroying links should probably occur through an `#update` on a paper unless we want to make links a resource in and of itself.

briankung:
See comment here: https://github.com/briankung/projectcredo/pull/153#issuecomment-261282247

briankung:
I'd recommend merging #153 into master and then merging or rebasing this branch off of master.

briankung (commented):
Woops, forgot to leave these comments. Damn this new system.

briankung:
I made the comment on #153, but I just wanted to clarify the strategy you could use to _not_ use ujs but still have a smooth user experience. I use something similar on my own personal site:

``` erb
<div>
  <%= link_to 'Destroy', done, method: :delete, data: { confirm: 'Are you sure?' } %>
  <b class="editable-entry"><%= done.text %></b>
  <%= render 'form', done: done, hidden: true %>
</div>
```

Per item, I'm pre-rendering a hidden form with that item's information in it already. When I want to edit it without reloading the page, I simply hide the content and unhide the form, which does not have to be done with ujs.

You _could_ optimize this even further and render only one template form somewhere that the js then grabs and populates with the item's data and inserting it in the content's place. Arguably this is slower on the client side, but faster on the server side since we're not rendering a hidden form per item.

wsf246:
Yea I had the same strategy for the tags but in getting concerned we are
pre rendering too many hidden things.  I feel like this is why u really
want that front end approach in place, this mix of ujs and JavaScript seem
hacky and not ideal

briankung:
Absolutely why I want a real front-end. That said in this case, the hidden
form would probably be preferable so as not to affect our backend routes as
much.

briankung:
Another thing to consider is that if we do migrate to a front end Library,
we will have to rewrite a lot of this logic, in all likelihood. That's why
I'd prefer to go light on all unnecessary JavaScript, at the cost of some
user experience.

briankung:
It's a prototype, so we should test our core assumptions before investing
too much time on UX that may not matter a whit.

wsf246:
I put links under papers in the routes, but did shallow: true, so it ends up effectively being the same for that destroy ujs.

app/assets/javascripts/references.js
```diff
@@ -4,6 +4,24 @@ $(document).ready(function() {
...
+  });
+
+  $(document).on('click', 'a.cancel-abstract-form', function(e) {
+    e.preventDefault();
+    abstractForm = $(e.target).closest("div.abstract-form");
+    abstractDiv = abstractForm.prev('div.edit-abstract');
+    abstractForm.toggleClass('hidden', true);
+    abstractDiv.toggleClass('hidden', false);
+  });
+
```
wsf246, line 24: I also decided to just render the forms for each reference for now.

briankung:
Damn, I should have used `shallow` elsewhere, as well, instead of `path: '/'`

briankung:
> it ends up effectively being the same for that destroy ujs.

One way it's not the same is that it requires an http delete (or the Rails approximation of it) instead of a get request, which I think was my main issue with the route before.

app/controllers/links_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+  def destroy
+    link = Link.find(link_params[:id])
+    url = link.url
+    paper = link.paper.title
+    link.destroy
+
+    reference = Reference.find(params[:reference])
+    respond_to do |format|
+      format.html { redirect_back(fallback_location: root_path, notice: "'#{url}' has been successfully removed from '#{paper}'") }
+      format.js { render('destroy_link.js.erb', locals: { reference: reference }) }
```
briankung: Since `destroy_link.js.erb` is in the `links` folder and is being referred to as its own resource, the common convention would just be to name this `destroy.js.erb`

app/controllers/references_controller.rb
```diff
@@ -1,9 +1,9 @@
 class ReferencesController < ApplicationController
   before_action :ensure_current_user, except: :show
   before_action :set_paper_locator, only: :create
+  before_action :set_reference, only: [:show, :edit_abstract]
```
briankung: Does `:edit_abstract` still exist?

app/models/services/pubmed/resource.rb
```diff
@@ -25,8 +25,13 @@ def mapper
         title:              lambda {|data| data.css('ArticleTitle').text },
         publication:        lambda {|data| data.css('Journal Title').text },
         doi:                lambda {|data| data.css('ArticleId[IdType=doi]').text },
-        pubmed_id:          lambda {|data| data.css('ArticleIdList ArticleId[IdType=pubmed]').text },
-        abstract:           lambda {|data| data.css('AbstractText').map(&:text).join("\n\n") },
+        pubmed_id:          lambda {|data| data.css('PMID').text },
+        abstract:           lambda do |data|
+          data.css('AbstractText').map do |a|
+            [a['Label'],a.text]
```
briankung: Why is this an array?

briankung (changes requested):
Really minor stuff now, looking good 👍

app/controllers/references_controller.rb
wsf246: Nope!

app/models/services/pubmed/resource.rb
wsf246: solid question.

briankung: As the join occurs at the end of the do/end block...I'm not  sure what the results of joining an array of arrays is.

briankung: Looks like it just joins everything:

``` ruby
[['whats','up'],['hey','there']].join ' '
#=> "whats up hey there"
```

That said, since this looks weird and is not obvious at first glance, we should probably change it.

briankung (approved):

app/models/services/pubmed/resource.rb
wsf246: so embarrassing fact, I don't actually know how to do multiple things in a block.  

Should I do  (a['Label'],a.text).join("\n\n") or put it in curly brackets?

briankung: I didn't realize, but you want two line breaks between the label and the text, as well?

PR #154, opened 2016-11-17
Permissions

### LINKED ISSUE #150: Users should be able to control who can edit their lists

briankung:
**Problem:** People should have some control over who can edit their Lists. Closes #150 

**Solution:** Add permission

**Complications:**

This is a lot of code. There are probably bugs, and I wrote in a lot of extra functionality that we're not using yet - visibility for instance. I'm also not guarding against adding comments, just direct edits to the list.

Todo:
- [x] Make remove contributor link work
- [x] Guard against comments (maybe?)
- [x] Clean up and refactor

**Feature Changelog:**

- Users can now add edit permissions to other users.

### DIFF
--- a/app/controllers/homepage_controller.rb
+++ b/app/controllers/homepage_controller.rb
@@ -1,8 +1,8 @@
 class HomepageController < ApplicationController
   def show
     if current_user
-      @pinned_lists = current_user.homepage.lists
-      @unpinned_lists = current_user.visible_lists.where.not(id: @pinned_lists.pluck(:id)).limit(9)
+      @pinned_lists = current_user.homepage.lists.uniq
+      @unpinned_lists = current_user.visible_lists.where.not(id: @pinned_lists.pluck(:id)).uniq.limit(9)
     else
       redirect_to lists_url
     end
--- a/app/controllers/homepage_controller.rb
+++ b/app/controllers/homepage_controller.rb
@@ -2,7 +2,7 @@ class HomepageController < ApplicationController
   def show
     if current_user
       @pinned_lists = current_user.homepage.lists.uniq
-      @unpinned_lists = current_user.visible_lists.where.not(id: @pinned_lists.pluck(:id)).uniq.limit(9)
+      @unpinned_lists = current_user.visible_lists.where.not(id: @pinned_lists.pluck(:id)).limit(9)
     else
       redirect_to lists_url
     end
--- a/app/controllers/homepage_controller.rb
+++ b/app/controllers/homepage_controller.rb
@@ -1,7 +1,8 @@
 class HomepageController < ApplicationController
   def show
     if current_user
-      @lists = current_user.homepage.lists
+      @pinned_lists = current_user.homepage.lists.default_sort
+      @unpinned_lists = List.where.not(id: @pinned_lists.pluck(:id)).limit(9).default_sort
     else
       redirect_to lists_url
     end
--- a/app/controllers/homepage_controller.rb
+++ b/app/controllers/homepage_controller.rb
@@ -1,8 +1,8 @@
 class HomepageController < ApplicationController
   def show
     if current_user
-      @pinned_lists = current_user.homepage.lists.default_sort
-      @unpinned_lists = current_user.visible_lists.where.not(id: @pinned_lists.pluck(:id)).limit(9).default_sort
+      @pinned_lists = current_user.homepage.lists
+      @unpinned_lists = current_user.visible_lists.where.not(id: @pinned_lists.pluck(:id)).limit(9)
     else
       redirect_to lists_url
     end
--- a/app/controllers/homepage_controller.rb
+++ b/app/controllers/homepage_controller.rb
@@ -2,7 +2,7 @@ class HomepageController < ApplicationController
   def show
     if current_user
       @pinned_lists = current_user.homepage.lists.default_sort
-      @unpinned_lists = List.where(visibility: :public).where.not(id: @pinned_lists.pluck(:id)).limit(9).default_sort
+      @unpinned_lists = current_user.visible_lists.where.not(id: @pinned_lists.pluck(:id)).limit(9).default_sort
     else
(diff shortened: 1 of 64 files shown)


briankung:
Review app here: https://afternoon-meadow-22528-pr-154.herokuapp.com/lists

briankung:
# As testuser

...you can add any other users in the system:
screen shot 2016-11-17 at 6 57 57 pm

# As a contributor, but not a moderator

...you can remove yourself from the list of contributors.
screen shot 2016-11-17 at 6 59 00 pm

wsf246:
I can't click into any of the lists from the homepage on the deployed heroku

briankung:
Well _obviously_ you didn't have permission to see them, so the whole site broke down

briankung:
All problems with not being logged in should be fixed. Permissions, authentication, and roles are a headache to manage. I'll probably want to refactor them into some sort of central place if it makes sense to do so

app/controllers/homepage_controller.rb
```diff
@@ -1,7 +1,8 @@
 class HomepageController < ApplicationController
   def show
     if current_user
-      @lists = current_user.homepage.lists
+      @pinned_lists = current_user.visible_lists.merge(current_user.homepage.lists)
+      @unpinned_lists = current_user.visible_lists.where.not(id: @pinned_lists.pluck(:id)).limit(9)
```
wsf246, line 5: _thumbs up_ for readability

wsf246:
Create new list seems to be acting up as well.

app/controllers/lists/members_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+    member = User.find params[:id]
+    removing_self = member == current_user
+
+    unless current_membership.can_moderate? || removing_self
+      return redirect_back(fallback_location: path, alert: 'You do not have permission to moderate this list')
+    end
+
+    list.members.delete member
+    if removing_self
+      return redirect_to(user_list_path(list.user, list), notice: "You have removed yourself from \"#{list.name}\"")
```
wsf246: I would make the language more clear that you're being removed as a contributor.  Probably "You have removed yourself as a contributor from"

app/controllers/users/lists_controller.rb
```diff
@@ -2,25 +2,31 @@ class Users::ListsController < ApplicationController
...
-    render 'lists/edit'
-  end
-
   def show
     @references = @list.references.joins(:paper).order(params_sort_order)
     render 'lists/show'
   end
 
+  def edit
+    render 'lists/edit'
```
wsf246, line 19: should we do any rejiggering of our views folders to make these fall in line with rails defaults, ie put this view in users/lists/edit?  I'm not sure that will even work.

I've also been thinking about how we should move the partials around, I'm always a little confused as to when to put a partial into papers or references folders.

app/models/homepage.rb
```diff
@@ -2,4 +2,8 @@ class Homepage < ApplicationRecord
   belongs_to :user
 
   has_and_belongs_to_many :lists
+
+  def lists *args
+    super(*args).uniq
```
wsf246, line 7: this is fun

app/views/lists/show.html.erb
```diff
@@ -1,9 +1,13 @@
...
-  created by <%= link_to @list.user.username, user_lists_path(@list.user) %>
-   - <%= link_to 'Edit', edit_user_list_path(@list.user, @list) %>
-  <% if @list.user == current_user %>
-     - <%= link_to 'Delete List', user_list_path(@list.user, @list), data: { confirm: 'Are you sure?' }, method: :delete %>
+  created by <%= link_to @list.owner.username, user_lists_path(@list.owner) %>
+  <% if current_user %>
+    <% if current_user.can_edit? @list %>
+     - <%= link_to 'Edit', edit_user_list_path(@list.owner, @list) %>
+   <% end %>
+    <% if @list.owner == current_user %>
```
wsf246, line 8: cool

config/routes.rb
```diff
@@ -10,6 +10,7 @@
   resources :lists, only: [:index, :new, :create] do
     resources :references, only: [:show, :create, :destroy]
     resource :vote, controller: 'lists/votes', only: [:create, :destroy]
+    resources :members, only: :destroy, controller: 'lists/members'
```
wsf246, line 15: members only lol

db/migrate/20161114214352_add_role_to_list_membership.rb
```diff
@@ -0,0 +1,5 @@
+class AddRoleToListMembership < ActiveRecord::Migration[5.0]
+  def change
+    add_column :list_memberships, :role, :integer, null: false, default: 10
```
wsf246, line 3: are you choosing integer because it's smaller?  Just curious.

wsf246:
So this doesn't affect adding or removing references yet right?

app/controllers/users/lists_controller.rb
briankung: As long as we're sharing the views, we can keep it like this. Otherwise we can move them into their respective files

app/models/homepage.rb
briankung: Yeah, a lot of the activerecord query methods were returning one list per listmembership, and there can be multiple memberships on one list...so I had to make sure they're unique.

db/migrate/20161114214352_add_role_to_list_membership.rb
briankung: Choosing integer because that's what you're supposed to do with enums

briankung:
Nope, does not yet affect references

db/migrate/20161114214352_add_role_to_list_membership.rb
briankung: But yes, it is probably more efficient to store a variety of states as integers and map them to strings in the application

PR #158, opened 2016-11-19
Validations and List Contributors

briankung:
**Problem:**

1. List moderators couldn't select a contributor when editing a list
2. Validations on listmemberships will break the app - not that we really have any

**Solution:** Invert the condition to allow selecting a contributor and build the listmembership before saving the list

**Complications:**

**Feature Changelog:**

- [bug] Allows selecting contributor while editing list
- [bug] Enable validations and fix associated bugs

### DIFF
--- a/app/views/lists/_form.html.erb
+++ b/app/views/lists/_form.html.erb
@@ -75,7 +75,7 @@
           members = if list.persisted? then @list.members else [current_user] end
           non_members = all_users - members
         %>
-        <%= f.select :members, non_members.map(&:username), {include_blank: 'None'}, {class: "form-control", disabled: current_user.can_moderate?(list)}  %>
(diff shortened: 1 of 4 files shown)


wsf246:
So it's intentional they can't add contributors while creating a new list right?

app/views/lists/_form.html.erb
```diff
@@ -75,7 +75,7 @@
           members = if list.persisted? then @list.members else [current_user] end
           non_members = all_users - members
         %>
-        <%= f.select :members, non_members.map(&:username), {include_blank: 'None'}, {class: "form-control", disabled: current_user.can_moderate?(list)}  %>
+        <%= f.select :members, non_members.map(&:username), {include_blank: 'None'}, {class: "form-control", disabled: !current_user.can_moderate?(list)}  %>
```
wsf246: This disables the creator from adding contributors on 'New List'

briankung: Dammit

briankung:
Should be good to go now 👍

PR #159, opened 2016-11-19
Prevents references from being added or destroyed if user cannot edit

briankung:
**Problem:** An unauthorized user could add and remove references to another user's list

**Solution:** Prevent unauthorized users from adding or removing references on another user's list

**Complications:** Not sure if I got all the cases

**Feature Changelog:**

- Prevent unauthorized users from adding or removing references on another user's list

### DIFF
--- a/app/controllers/references_controller.rb
+++ b/app/controllers/references_controller.rb
@@ -8,8 +8,13 @@ class ReferencesController < ApplicationController
 
   def create
     list = List.find(reference_params[:list_id])
+    list_path = user_list_path(list.owner, list)
 
-    return redirect_to(:back, alert: @locator.errors.join(' ')) unless @locator.valid?
+    unless current_user.can_edit? list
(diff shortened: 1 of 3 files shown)


wsf246:
what do you think about removing the search bar and the 'add reference' forms if they are not contributors.  Or disabling them.

briankung:
Probably a good idea

wsf246:
If this is good, I'm gonna merge.  We can disable forms and the such on a separate PR.

briankung:
Good for now. I left my laptop at home by accident and I'm out. Feel free
to pull and disable the forms yourself. Just push up to the PR branch

briankung:
Or we could do it in a separate PR, up to you. I won't have my laptop for
another hour at least

wsf246:
you do you man, I'll merge for now.

briankung:
We also need to limit who can delete comments. Probably only a moderator?

wsf246:
agreed

PR #160, opened 2016-11-19
Requires moderator permissions to delete comments

briankung:
**Problem:** Non-moderator users can delete comments

**Solution:** Prevent them from doing so

**Complications:** Caught a bunch of other minor things, too. Ended up screwing up a rebase, too, so I'm taking this opportunity to try squashing my commit history as well.

**Feature Changelog:**

- Prevents a user from deleting comments or adding contributors unless they're a moderator
- Disables paper search bar and add paper unless the user can edit the list
- Prevents a user from adding an abstract unless they're logged in
- Allows a user to moderate their own comment
- Prevents moderator from editing others' comments

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -51,7 +51,15 @@ class CommentsController < ApplicationController
   # DELETE /comments/1.json
   def destroy
     reference = @comment.root.commentable
+    list = reference.list
+
+    unless current_user.can_moderate?(list)
+      flash[:alert] = "You do not have permission to moderate this list"
+      return redirect_back(fallback_location: user_list_path(list.owner, list))
+    end
+
     @comment.destroy
+
     respond_to do |format|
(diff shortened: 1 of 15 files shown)


wsf246:
dunno if this happened on this PR or another, but it looks like I can't delete my own comment.  I can edit my own comment though.  I would think we should let people delete their own comment even if they are not contributors/moderators.

briankung:
It's probably from this PR

wsf246:
the "edit" link oddly dissapears after I edit a comment.

Also, I can still edit and update an abstract, add a link, and add tags if im not a contributor for a list.  I'm kinda ok with this.

briankung:
Eh, different PR

briankung:
This one is already expanding too far

briankung:
Jk, fixed the edit weirdness. The abstract, link, and tags I'm cool with, as well

PR #161, opened 2016-11-20
Only shows lists that the params user owns

briankung:
**Problem:**
Routes to lists were resolving for any user who had a membership, so `briankung/maintaining-mobility-in-old-age` lead to the same list as `testuser/maintaining-mobility-in-old-age`

**Solution:** Limit `users/lists#show` to lists that the user actually owns

**Complications:** I'm sure this will be a continuing headache to manage

**Feature Changelog:**

- Limit showing lists that the user in the url params actually owns

### DIFF
--- a/app/views/lists/show.html.erb
+++ b/app/views/lists/show.html.erb
@@ -1,6 +1,6 @@
 <h2 class= "list-title"><%= @list.name %></h2>
 <div class="list-nav">
-  created by <%= link_to @list.owner.username, user_lists_path(@list.owner) %>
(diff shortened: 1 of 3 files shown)


app/views/lists/show.html.erb
```diff
@@ -1,6 +1,6 @@
 <h2 class= "list-title"><%= @list.name %></h2>
 <div class="list-nav">
-  created by <%= link_to @list.owner.username, user_lists_path(@list.owner) %>
+  by <%= link_to @list.owner.username, user_lists_path(@list.owner) %>
```
wsf246, line 3: _thumbs up_

PR #163, opened 2016-11-21
Simplifies resources

briankung:
**Problem:** Bugs and refactoring

**Solution:** Fix the bugs, refactor

**Complications:** This was in the hopes of implementing limited link scraping, but it ended up not being worth it in my opinion:

> We need to know what links we need to be able to scrape before we can decide which ones to scrape or how to do it

> so basically...maybe we let people input crap data for a little bit until we know what to prioritize

**Feature Changelog:**

- Bugfix: Guards against nil labels
- Bugfix: Guards against empty PubmedArticleSet for locator_ids like 1234567890987654321

### DIFF
--- a/app/models/services/crossref/resource.rb
+++ b/app/models/services/crossref/resource.rb
@@ -1,24 +1,23 @@
 class Crossref
   class Resource
-    attr_accessor :id, :response
+    attr_accessor :id, :response, :paper_attributes
 
     def initialize id
       self.id = id.to_s
       endpoint = URI.parse("https://api.crossref.org/works/#{self.id}")
       self.response = Net::HTTP.get_response endpoint
-    end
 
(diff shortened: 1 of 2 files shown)


app/models/services/crossref/resource.rb
```diff
@@ -1,24 +1,23 @@
 class Crossref
   class Resource
-    attr_accessor :id, :response
+    attr_accessor :id, :response, :paper_attributes
```
wsf246, line 3: sweet

wsf246:
This is great.  One thing that we can either do here or in another PR, I've run into some of the same publishing date issues that we've had with crossref as well.  Otherwise LGTM.

briankung:
I'm not sure what the crossref issues are, so let's save that for another pr

PR #164, opened 2016-11-21
Adds import source to papers for future data purposes

briankung:
**Problem:** We're going to need to be able to differentiate Papers by their import source in order to fix Papers that were imported by link.

**Solution:** Note a paper's import source

**Complications:** `¯\_(ツ)_/¯`

### DIFF
--- a/app/models/services/crossref/resource.rb
+++ b/app/models/services/crossref/resource.rb
@@ -28,6 +28,7 @@ class Crossref
 
     def mapper
       {
+        import_source:      lambda {|data| 'crossref' },
         title:              lambda {|data| data.dig 'message', 'title', 0 },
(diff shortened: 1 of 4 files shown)


wsf246:
I think gets back at the question of whether we're mixing API data.  If the paper itself ends up being a mix of content (found through crossref but gets abstract from pubmed), we should probably put this attribute on the API import responses instead.

wsf246:
I have a "paper_editable" attribute as well.  My plan was to make anything that came through URL "paper editable".

briankung:
I think this has the connotation of "original import source." I don't think import source would change if you were to add an abstract, for example.

briankung:
I'm going to merge.

PR #166, opened 2016-11-21
Public by default

briankung:
**Problem:** There was no way to accept list contributions from anyone

**Solution:** Turn on public contributions by default

**Complications:** Authorization is getting messy. I also found a bug with `remote: true` delete links.

**Feature Changelog:**

- Turn on public contributions to lists by default
- Allow users to select who can contribute to a list


wsf246:
I think the abstract conditional if there's no abstract is off:

image

I'm signed in and getting the wrong message.

PR #184, opened 2016-11-27
Signin fix

briankung:
**Problem:** Post-sign in would break

**Solution:** Use correct path

**Complications:**

**Feature Changelog:**

- [bug] Site no longer breaks after log in


wsf246:
Jesus that's a lot of commits to make sure the site doesn't break on sign
in.

wsf246:
I'm confused as to which version we're supposed to use to log issues.
 should i use brian kung's or projectcredo's?  Also, we should prob route
the github link in the about page to the correct one.

briankung:
I opened this on the wrong base branch. See projectcredo/projectcredo#9

Yes, and link to the tutorial video as well

